PR #3, opened 2018-11-29
React redux refactor

gkallenberg:
Large refactor for handling store across components via Redux and linking an action client-side for calling the search API via axios. This pull request is more of a proof of concept for adding basic Redux logic for handling state.

Includes new components for handling results lists in general. Design elements come from NYPL's Design Toolkit for now and will change once UX team integration happens.

Server-side integration is lacking.

Unit tests on new and existing components are needed as well as documentation.


.env.sample
```diff
@@ -0,0 +1,4 @@
+NYPL_API_BASE_URL
```
EdwinGuzman: Do you need something like `NYPL_API_BASE_URL=example/URL`?

src/app/actions/Actions.js
```diff
@@ -1,18 +1,11 @@
...
-  }
-
-  searchByAuthor(data) {
-    return data;
-  }
-}
-
-export default alt.createActions(Actions);
+// Actions
+export const Actions = {
```
EdwinGuzman: _I think_ it would be okay to have both the constants and the actual action functions in the same class or file.

src/app/actions/SearchActions.js
```diff
@@ -0,0 +1,33 @@
...
+
+export const search = (query, filter = 'q') => {
+  // Need a parsed query input to use for each filter
+  const userQuery = (query) ? encodeURI(query) : 'test';
+  // Need a client to send the search and receive results
+  // Need to pass the results to a renderer
+  const apiUrl = appConfig.esUrl.development;
+
+  return (dispatch) => {
+    return axios.get(apiUrl, { params: { q: userQuery } })
```
EdwinGuzman: You mentioned that in your function you are not returning a promise like I am in my branch, but this axios call is a promise so they're both alike.

src/app/components/Application/Application.jsx
```diff
@@ -3,13 +3,13 @@ import React from 'react';
...
-import Store from '../../stores/Store';
+import SearchResultsList from '../Search/SearchResultsList';
 
-class App extends React.Component {
+class Application extends React.Component {
   constructor(props) {
     super(props);
 
-    this.state = Store.getState();
+    this.state = props;
```
EdwinGuzman: I don't believe you will need this.

src/app/components/Search/Instances.jsx
```diff
@@ -0,0 +1,66 @@
...
+  render() {
+    const items = this.parseInstances(this.instances);
+
+    return (
+      <ul className="nypl-instances-list">
+        {
+          items.map((element, key) => (
+            <li className="nypl-results-item">
+              <div className="nypl-results-description">
+                <p tabIndex={key.toString()} key={key.toString()}>
```
EdwinGuzman: You don't and shouldn't define `tabIndex` like this. Typically we try not to give element tab indices unless if it's "0" or "-1".

src/app/components/Search/Instances.jsx
```diff
@@ -0,0 +1,66 @@
...
+  };
+
+  render() {
+    const items = this.parseInstances(this.instances);
+
+    return (
+      <ul className="nypl-instances-list">
+        {
+          items.map((element, key) => (
+            <li className="nypl-results-item">
```
EdwinGuzman: The top level element inside the map function (or any looped function) is the one that needs the `key={..}` attribute.

src/app/components/Search/ResultsList.jsx
```diff
@@ -0,0 +1,36 @@
...
+    return null;
+  }
+
+  return (
+    <div>
+      <h2>Works</h2>
+      <ul className="nypl-results-list">
+        {
+          results.results.map((result, i) => (
+            <li className="nypl-results-item" tabIndex={i.toString()} key={i.toString()}>
```
EdwinGuzman: No `tabIndex`.

src/app/components/Search/ResultsMetadata.jsx
```diff
@@ -0,0 +1,19 @@
...
+  if (metadata.metadata.total === 0) {
+    return (
+      <div className="nypl-results-summary" aria-live="assertive" aria-atomic="true" role="presentation">
+        Your search yielded no results. Please try again.
+      </div>
+    );
+  }
+
+  return (
+    <div className="nypl-results-summary" aria-live="assertive" aria-atomic="true" role="presentation">
```
EdwinGuzman: Does this work with screenreaders? You can try always rendering an empty div with the `aria-live` attribute and just updating the text based on `metadata.metadata.total`.

src/app/components/Search/SearchQuery.js
```diff
@@ -0,0 +1,29 @@
+import axios from 'axios';
```
EdwinGuzman: This file isn't being used.

src/app/components/Search/SearchResultsList.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+class SearchResultsList extends React.Component {
+  render () {
+    const metadata = this.props.results && this.props.results.hits;
+    const hits = this.props.results && this.props.results.hits && this.props.results.hits.hits;
+
+    let resultsList;
+
+    if (!hits) {
+      resultsList = null;
```
EdwinGuzman: You could also `return null;` here instead of always rendering an empty div.

src/app/stores/Store.js
```diff
@@ -1,20 +1,32 @@
-import Actions from '../actions/Actions.js';
-import alt from '../alt.js';
+import Actions from '../actions/Actions';
+import alt from '../alt';
```
EdwinGuzman: Won't need this Alt store anymore!

src/server/ApiRoutes/ApiRoutes.js
```diff
@@ -1,40 +1,26 @@
...
-  //   .catch(error => {
-  //     // Do something with the error
-  //     res.locals.data = {
-  //       Store: {},
-  //     };
-  //
-  //     next();
-  //   });
-
   res.locals.data = {
```
EdwinGuzman: This won't be needed anymore. There will probably be a different way of sending data from the server to the client. Redux calls it "hydrating" but not sure what it involves yet.

EdwinGuzman:
Some minor changes are needed. Overall it's good and works. I didn't look closely at the CSS which can be improved but I imagine that will be changing anyway.

src/app/actions/SearchActions.js
```diff
@@ -0,0 +1,33 @@
...
+
+export const searchResults = (results) => {
+  return {
+    type: Actions.SEARCH,
+    results,
+  };
+};
+
+export const search = (query, filter = 'q') => {
+  // Need a parsed query input to use for each filter
```
mwbenowitz: Are "filters" Elastic filters or a redux component? If they're for Elastic you shouldn't need them to make queries. Also if we intended to use filters we'll probably need to POST a more complex JSON block to Elastic, the URI search doesn't expose all of the Elastic functionality

src/app/actions/SearchActions.js
```diff
@@ -0,0 +1,33 @@
...
+  const userQuery = (query) ? encodeURI(query) : 'test';
+  // Need a client to send the search and receive results
+  // Need to pass the results to a renderer
+  const apiUrl = appConfig.esUrl.development;
```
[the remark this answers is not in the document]
EdwinGuzman: Yup, wouldn't say it's absolutely necessary right now though.

.env.sample
gkallenberg: The next iteration should include integration with the API Gateway which will use this variable. For now, the endpoint is not fully implemented and integrated with the gateway.

src/app/actions/SearchActions.js
gkallenberg: My thinking was that `filters` would be used to match fields in Elastic and I believe should to be captured in a store for keeping track of search history for proper unfiltering and browser history storage for going back via the browser's Back button. I know that tracking history can be tricky, but perhaps Redux can help with this.

We'll definitely need to work on how we're sending queries from the front end to the API and formulating the proper query for sending along to Elastic to take advantage of Elastic's features. Additional work will need to be done on the Search API to parse what's being sent from the front end, send the proper query, and handle the response properly for delivery back to the front end

I'll need to do some research on this, and I know Paul has wrangled with query handling and it can be problematic to get right.

src/app/actions/SearchActions.js
gkallenberg: Right -- I was just wondering if use of `new Promise(resolve, reject)` with an internal call to the search function separately was a better approach than just calling axios here.

src/app/components/Application/Application.jsx
gkallenberg: I'm cutting this out.

src/app/components/Search/Instances.jsx
gkallenberg: `tabIndex` is being removed

src/app/components/Search/Instances.jsx
gkallenberg: adding the key

src/app/components/Search/ResultsList.jsx
gkallenberg: removed

src/app/components/Search/ResultsMetadata.jsx
gkallenberg: just added this from the design toolkit recommendations. I'm sure that's not necessarily the way to go and all of this may change when UX gets more involved.

src/app/stores/Store.js
gkallenberg: removed

src/app/actions/SearchActions.js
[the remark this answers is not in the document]
gkallenberg: Changed this to use process.env value set in webpack.config.js

README.md
```diff
@@ -1,19 +1,49 @@
...
+
+### Dependencies
+
+* ES6 - compiled with Babel and Webpack
 * Redux
+
+### Dev Dependencies
+
 * Webpack-dev-server
 * NYPL Header and Footer npm modules
```
EdwinGuzman: These are not dev-dependencies.

README.md
```diff
@@ -1,19 +1,49 @@
...
 
 ### Usage
-Clone the repo and run `npm install`.
 
-Then run `npm start` to start the local server at `localhost:3001`.
+Currently takes in a query string to pass along to the ResearchNow Search API which submits a keyword search to the Elasticsearch instance, and renders the returned output. Sends the `q` parameter specified in the rendered search form on the main page.
+
+The server route `/search` should also take a `q` query string parameter and perform the same keyword search via the ResearchNow Search API.
 
+Future functionality will take a title and author field for more preceise searching.
```
EdwinGuzman: Precise*

src/app/components/Search/Instances.jsx
```diff
@@ -1,66 +1,50 @@
 import React from 'react';
```
EdwinGuzman: Is "Instances" related to elasticsearch?

src/app/components/Search/Instances.jsx
```diff
@@ -1,66 +1,50 @@
...
-
-    this.instances = props;
+export const Instances = (instances) => {
+  if (_isEmpty(instances.instances)) {
+    return null;
   }
 
-  parseInstances(instances) {
-    const items = instances.instances.map((instance, key) => (
+  const items = instances.instances.map((instance, key) => (
```
EdwinGuzman: You don't use `key` here so you don't need it.

src/app/components/Search/Instances.jsx
```diff
@@ -1,66 +1,50 @@
...
       {
-        ebooks: (instance.items) ? instance.items : [],
-        pub_date: (instance.pub_date) ? parseInt(instance.pub_date) : null,
-        pub_place: (instance.pub_place) ? instance.pub_place : null,
-        publisher: (instance.publisher) ? instance.pub_publisher : null,
-        language: (instance.language) ? instance.language : null
+        items.map((element, key) => (
+          <li className="nypl-results-item" key={key.toString()}>
+            <div className="nypl-results-description">
+              <p key={key.toString()}>
```
EdwinGuzman: You don't need the `key` attribute here since it's already in the preferred `li` element.

src/app/components/Search/Instances.jsx
```diff
@@ -1,66 +1,50 @@
...
+                {
+                  element.ebooks.map((ebook, ebookKey) => (
```
EdwinGuzman: Will `element` always contain `ebooks`? Because this may not work if `element.ebooks` is undefined. A check to make sure it's defined would be helpful.

EdwinGuzman: But I guess you did default it to `[]` so that's good.

src/app/components/Search/ResultsList.jsx
```diff
@@ -4,8 +4,7 @@ import {
 } from 'underscore';
 import Instances from './Instances';
 
-const ResultsList = (results) => {
-  console.log(results);
+export const ResultsList = (results) => {
```
EdwinGuzman: You can also `export default const ...` so when you import you can do `import ResultsList from '...'` instead of `import { ResultsList } from '...'`.

src/app/search/query.js
```diff
@@ -1,8 +1,8 @@
...
-  const q = query.q || '';
+export const getRequestParams = (query = {}) => {
+  const q = query.q || 'test';
   const filters = query.filters || {};
 
   return { q, filters };
-}
+};
 
 export default getRequestParams;
```
EdwinGuzman: You're exporting this twice and that's not necessary.

src/server/ApiRoutes/Search.js
```diff
@@ -1,53 +1,12 @@
...
-      console.log('Error returning results server side.', error);
-      next();
-    },
-  );
-}
+  next();
+};
 
-export default { searchServer, search };
+export default { searchServer };
```
EdwinGuzman: Exporting twice again.

test/Instances.test.js
```diff
@@ -0,0 +1,40 @@
...
+          pub_place: '',
+          publisher: '',
+          language: '',
+        },
+      ];
+      component = shallow(<Instances instances={instances} />);
+    });
+
+    it('should return null if empty.', () => {
+      expect(component.find('.nypl-instances-list')).to.have.length(1);
```
EdwinGuzman: This test is fine but in reality, would you want to have a list of an empty result?

src/app/actions/SearchActions.js
EdwinGuzman: Leave this as is and refactor later if you find that's helpful.

src/app/components/Search/ResultsMetadata.jsx
EdwinGuzman: Cool. I think leaving it as is is fine.

src/app/components/Search/SearchForm.jsx
```diff
@@ -1,56 +1,132 @@
...
+import SearchButton from '../Button/SearchButton';
+import { search } from '../../actions/SearchActions';
 
-class SearchForm extends React.Component {
+export class SearchForm extends React.Component {
   constructor(props) {
     super(props);
 
-    this.state = Store.getState();
+    this.state = props;
```
EdwinGuzman: You can remove this from here.

src/app/components/Search/SearchForm.jsx
```diff
@@ -1,56 +1,132 @@
...
+    event.preventDefault();
+    if (!this.state.filter) {
+      this.setState({ filter: 'q' });
+    }
+
+    this.props.search(this.state.query, this.state.filter);
   }
 
   render() {
+    const { searchResults, query, filter, allowedFilters, sort } = this.state;
```
EdwinGuzman: You're not using `searchResults` in this component so it might be good to remove here (and any other variable that's not being used).

src/app/components/Search/Instances.jsx
gkallenberg: Instances relates to the data structure language used for describing items which would be the equivalent in Sierra.

src/app/actions/SearchActions.js
mwbenowitz: Got it, that makes sense. I was a little confused because I see `filters` as more of an advanced search/browse functionality, but of course it should be supported here. I think our work on the API will help inform this in the next iterations, especially since it will provide at least _some_ constraints on how we pass data back to Elastic through the API

src/app/components/Search/ResultsList.jsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <div>
+      <h2>Works</h2>
+      <ul className="nypl-results-list">
+        {
+          results.results.map((result, i) => (
+            <li className="nypl-results-item" key={i.toString()}>
+              <h3>
+                <a href={`/work/detail/${result._id}`}>
+                  {result['_source'].title} &ndash; {result['_source'].entities[0].name}
```
mwbenowitz: Just leaving a note that as we work on the API the response block will become easier to parse/more logical, so we won't have to deal with the nested `data` structure and the underscore variable names.

EdwinGuzman: A modeling function can be created to create the data structure _for_ the front-end. Whenever the data structure in the backend gets updated, you'd only have to update that one modeling function to update the data structure for the whole app.
Perhaps something along the lines of  https://github.com/NYPL-discovery/discovery-front-end/blob/development/src/app/utils/item.js#L45

gkallenberg:
Thanks, all. Did some minor edits for removing unused/unnecessary variable/keys and duplicate export statements.

EdwinGuzman:
Closing this because @gkallenberg based of merged commits out of this PR and it's no longer needed!

src/app/components/Search/SearchForm.jsx
```diff
@@ -1,56 +1,132 @@
...
+                  >
+                    <option value={allowedFilters.kw} defaultValue>
+                      Keyword
+                    </option>
+                    <option value={allowedFilters.ti}>
+                      Title
+                    </option>
+                    <option value={allowedFilters.au}>
+                      Author
+                    </option>
```
mwbenowitz: Sorry to get hung up on this but I think these "filters" are better termed "queries/search queries" since that's the block they'll end up getting passed to in Elastic. Filters will be a separate block in the API. Just trying to anticipate variables that might end up being confusing.

PR #4, opened 2018-12-07
Summary results refinement

gkallenberg:
This is a relatively minor set of changes that address the summary view requirements of [SFR-168](https://jira.nypl.org/browse/SFR-168).

* Fixes the variable outputting the publisher since there was a copy/paste issue.
* Changes references from filter to field since this language is more in line with Elasticsearch's language.
* Moved Instances to ResultsRow in order to be more generic about the type of object we're processing in results pages. 
* Added fixtures for result sets and use them for the results component tests.


src/app/components/Search/ResultsList.jsx
```diff
@@ -17,12 +17,12 @@ const ResultsList = (results) => {
           results.results.map((result, i) => (
             <li className="nypl-results-item" key={i.toString()}>
               <h3>
-                <a href={`/work/detail/${result._id}`}>
+                <a href={`/work/${result._id}`}>
```
EdwinGuzman: This is good but you could also try to use react-router's `<Link>` component to link to different states in the app.

EdwinGuzman: Not needed for this PR.

src/app/components/Search/SearchForm.jsx
```diff
@@ -115,18 +117,18 @@ SearchForm.defaultProps = {
...
 };
 
 const mapDispatchToProps = dispatch => ({
-  search: (query, filter) => dispatch(search(query, filter)),
+  search: (query, field) => dispatch(search(query, field)),
 });
 
 
-export default connect(
+export default withRouter(connect(
```
EdwinGuzman: Never used `withRouter` but will agree with what you do.

src/app/components/Search/ResultsList.jsx
gkallenberg: reverting this change and will make adjustments in the next PR which will be work on the item pages.

src/app/components/Search/ResultsRow.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <ul className="nypl-rows-list">
+      {
+        items.map((element, key) => (
+          <li className="nypl-results-item" key={key.toString()}>
+            <div className="nypl-results-description">
+              <p>
+                {
+                  element.ebooks.map((ebook, ebookKey) => (
```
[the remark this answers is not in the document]
gkallenberg: added

src/app/components/Search/SearchForm.jsx
gkallenberg: Am going to move this to the next PR since I'm prepping for adding changes to server.js, etc. for server-side rendering and routing. It was suggested in the React Router docs to use this when implementing Redux on a project.

src/app/components/Search/ResultsRow.jsx
```diff
@@ -0,0 +1,52 @@
...
+    }
+  ));
+
+  return (
+    <ul className="nypl-rows-list">
+      {
+        items.map((element, key) => (
+          <li className="nypl-results-item" key={key.toString()}>
+            <div className="nypl-results-description">
+              <p>
```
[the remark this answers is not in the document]
gkallenberg: changing to a list

PR #5, opened 2019-01-15
Container refactor

gkallenberg:
Moved some of the connect logic to a Search container for passing props and actions down to child components who in turn can summon actions from the container. Simplifies the child components which receive props instead of trying to manage state/storage in addition to the higher level components.

I think it's a bit cleaner and seems to be provide a clearer path to adding the details page and/or other child components.

Still working on the server side rendering piece as well, and this may help with that. Am still learning.


src/app/components/Search/Search.jsx
```diff
@@ -0,0 +1,58 @@
...
+import { bindActionCreators } from 'redux';
+import SearchForm from './SearchForm';
+import SearchResultsList from './SearchResultsList';
+import * as searchActions from '../../actions/SearchActions';
+
+class Search extends React.Component {
+  constructor(props) {
+    super(props);
+    const { dispatch } = props;
+    this.state = props;
```
EdwinGuzman: You're not using the state anywhere  in this function it seems? Unless if you need to initialize it here so that the `mapStateToProps` function works?

src/app/components/Search/ResultsList.jsx
```diff
@@ -1,30 +1,29 @@
...
     <div>
       <h2>Works</h2>
       <ul className="nypl-results-list">
-        {
-          results.results.map((result, i) => (
+        {props.results.map((result, i) => (
             <li className="nypl-results-item" key={i.toString()}>
               <h3>
-                <a href={`/work/detail/${result._id}`}>
+                <Link to={{ pathname: '/work', query: { workId: `${result['_source'].ids[0].identifier}` } }}>
```
mwbenowitz: It may be easier/safer to have this lookup works by UUID, which is always present and exists outside of the identifier array (and is guaranteed to be unique. While identifiers should uniquely identify works, I'm not 100% we can trust our sources).

If the single work endpoint doesn't support this right now I could potentially add that.

src/app/actions/SearchActions.js
```diff
@@ -18,20 +18,21 @@ export const search = (query, field = 'q') => {
...
       .then((resp) => {
         if (resp.data) {
           dispatch(searchResults(resp.data));
         }
       })
       .catch((error) => {
-        console.log('Error communicating with Elasticsearch', apiUrl, error);
-        throw new Error('Error communicating with Elasticsearch', apiUrl, error);
+        console.log('Error communicating with Elasticsearch', esUrl, userQuery, error);
+        throw new Error('Error communicating with Elasticsearch', esUrl, userQuery, error);
```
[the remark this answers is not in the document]
EdwinGuzman: I think it's the same url just different name?

src/app/components/Search/ResultsList.jsx
EdwinGuzman: This seems to be placeholder because the routes aren't fully set up yet. But I do agree about the UUID.

src/app/actions/SearchActions.js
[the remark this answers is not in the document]
mwbenowitz: It looks like esUrl is the bare beanstalk endpoint that the apiUrl serves as a proxy to, just wondering what the reason was for removing the Gateway url, it should make no difference functionally.

src/app/components/Search/Search.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import SearchForm from './SearchForm';
+import SearchResultsList from './SearchResultsList';
+import * as searchActions from '../../actions/SearchActions';
+
+class Search extends React.Component {
```
EdwinGuzman: From the architecture of this app and component, I think maybe `SearchContainer` would be a better name for this container. I was expecting to just find the search form in `Search` but it turns out to be the layout for the main page.

src/app/components/Search/Search.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
```
EdwinGuzman: I'm not sure if this is the best place to put this. You might be able to make this more generic if it was outside of this component but I also don't know exactly how this function works.

EdwinGuzman:
You can take out the search result list related components from the `Search` folder and move them to their own folder to group related components together.

The `ResultList` is a list of results, and each results renders a `Results Row` which is a list of items. Each item also has a list and that's a list of ebooks. Right? Just trying to make sense of it.

The refactor and making the `Search` container connect to the store and passing down props to the children looks good.

src/app/components/Search/Search.jsx
gkallenberg: Removing this line as it is superfluous

src/app/components/Search/Search.jsx
gkallenberg: Will rename. Wasn't sure if this was actually going to work and wasn't sure how specific to be in naming for container, component, etc.

src/app/components/Search/Search.jsx
gkallenberg: There might be a better home for this at the Application level perhaps. Not sure exactly what best practices are for Redux based React apps, but can certainly work on refactoring more as more experience is gleaned.

EdwinGuzman: Leave it for now. We can update later as we learn more.

src/app/components/Search/ResultsList.jsx
gkallenberg: Yes, a standard ID should be used -- definitely placeholder code to basically get used to using the <Link/> component of react-router. Will change to use the UUID property instead.

gkallenberg:
Moving components to respective group folders. My thinking was that the SearchContainer would control the actions/data flow and contain its related children. There will probably be a container for filters as well. At least that's my initial thinking since a FilterContainer may need to manage its own data and communicate back to the SearchContainer.

The ResultsList component relates the top-level bib record and the ResultsRow component displays each instance or item or work that is associated with that bib record. Basically the ResultsRow is displaying different bib records of editions related to the top-level bib record displayed by the ResultsList markup. Each row can have eBook links, as well as bibliographic detail related to that row's edition. Grouping occurs when each edition, format, etc. of a work relates to a top-level or shared bibliographic record.

I'm responsible for creating some architectural diagrams of the components as well as documentation describing what we're trying to achieve. I'm finding that more planning is necessary to make sure that the containers and components all make sense and can be flexible enough to accommodate change.

EdwinGuzman:
Cool. That's a great write up and hope to see the architectural diagrams soon.

src/app/components/SearchForm/SearchForm.jsx
```diff
@@ -0,0 +1,113 @@
...
+                      </option>
+                      <option value={this.props.allowedFields.au}>
+                        Author
+                      </option>
+                    </select>
+                  </span>
+                </div>
+                <div className="nypl-text-field">
+                  <span className="nypl-omni-fields-text">
+                    {/* <label htmlFor="search-input">Search ResearchNow for</label> */}
```
EdwinGuzman: Oh, I think you do need this label element eventually.

gkallenberg: Yeah -- I commented out because it was causing display issues, but left it in because it's needed.

PR #6, opened 2019-01-17
Add WorkDetail Page

gkallenberg:
Adds new actions and client-side logic to fetch individual records from Elasticsearch by UUID and display a work's details.

Adds an onClick action to each search result's title link to fetch and display more details for that particular work and its editions.

### DIFF
--- a/src/app/actions/SearchActions.js
+++ b/src/app/actions/SearchActions.js
@@ -68,8 +68,6 @@ export const fetchWork = (workId) => {
       .then((resp) => {
         if (resp.data) {
           dispatch(workDetail(resp.data));
-          // Do I need to change state of the searchResults here?
-          dispatch(searchResults());
         }
       })
       .catch((error) => {
--- a/src/app/actions/SearchActions.js
+++ b/src/app/actions/SearchActions.js
@@ -3,6 +3,7 @@ import appConfig from '../../../appConfig';
 
 export const Actions = {
   SEARCH: 'SEARCH',
+  FETCH_WORK: 'FETCH_WORK',
 };
 
 export const searchResults = (results) => {
@@ -12,26 +13,70 @@ export const searchResults = (results) => {
   };
 };
 
-export const search = (query, field = 'q') => {
-  const appEnv = process.env.APP_ENV || 'production';
+export const workDetail = (item) => {
+  return {
+    type: Actions.FETCH_WORK,
+    item,
+  };
+};
+
+const appEnv = process.env.APP_ENV || 'production';
+const apiUrl = appConfig.api[appEnv];
+const { searchPath, recordPath } = appConfig.api;
+const searchUrl = apiUrl + searchPath;
+
+export const searchGet = (query) => {
   // Need a parsed query input to use for each filter
   const userQuery = (query) ? encodeURIComponent(query) : '*';
-  // Need a client to send the search and receive results
-  // Need to pass the results to a renderer
-  const apiUrl = appConfig.api[appEnv];
 
   return (dispatch) => {
-    return axios.get(apiUrl, { params: { q: userQuery } })
+    return axios.get(searchUrl, { params: { q: userQuery } })
       .then((resp) => {
         if (resp.data) {
           dispatch(searchResults(resp.data));
         }
       })
       .catch((error) => {
-        console.log('Error communicating with Elasticsearch', apiUrl, userQuery, error);
-        throw new Error('Error communicating with Elasticsearch', apiUrl, userQuery, error);
+        console.log('An error occurred during searchGet', error);
+        throw new Error('An error occurred during searchGet', error);
       });
   };
 };
 
-export default { search };
+export const searchPost = (query, field) => {
+  // Need a parsed query input to use for each filter
+  const userQuery = (query) ? encodeURIComponent(query) : '*';
+
+  return (dispatch) => {
+    return axios.post(searchUrl, { queries: [{ field, value: userQuery }] })
+      .then((resp) => {
+        if (resp.data) {
+          dispatch(searchResults(resp.data));
+        }
+      })
+      .catch((error) => {
+        console.log('An error occurred during searchPost', error);
+        throw new Error('An error occurred during searchPost', error);
+      });
+  };
+};
+
+export const fetchWork = (workId) => {
+  const recordUrl = apiUrl + recordPath;
+  return (dispatch) => {
+    return axios.get(recordUrl, { params: { recordID: workId } })
+      .then((resp) => {
+        if (resp.data) {
+          dispatch(workDetail(resp.data));
+          // Do I need to change state of the searchResults here?
+          dispatch(searchResults());
+        }
+      })
+      .catch((error) => {
+        console.log('An error occurred during fetchWork', error);
+        throw new Error('An error occurred during fetchWork', error);
+      });
+  };
+};
+
+export default { searchGet, searchPost, fetchWork };
--- a/src/app/components/SearchContainer/SearchContainer.jsx
+++ b/src/app/components/SearchContainer/SearchContainer.jsx
@@ -2,9 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
-import {
-  isEmpty as _isEmpty,
-} from 'underscore';
 import SearchForm from '../SearchForm/SearchForm';
 import SearchResults from '../SearchResults/SearchResults';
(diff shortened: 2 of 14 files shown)


src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -2,8 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
+import {
+  isEmpty as _isEmpty,
+} from 'underscore';
```
mwbenowitz: Looks like this is unused?

src/app/components/WorkDetail/WorkDetail.jsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const { item } = props.detail;
+  const ebookUrl = appConfig.ereader[process.env.APP_ENV];
+
+  const parseEntities = (entities) => {
+    return (
+      <ul>
+        {entities.map((entity, i) => (
+          <li key={i.toString()}>
+            <Link to={{ pathname: '/', query: { author: `${entity.name}` } }}>{entity.name}, {entity.role}</Link>
```
mwbenowitz: It is hypothetically possible that an entity will have multiple roles per work (e.g. author, editor). I'm not sure if any of the test records have this, but I have seen it in Gutenberg/OCLC data

src/app/actions/SearchActions.js
```diff
@@ -12,26 +13,70 @@ export const searchResults = (results) => {
...
+};
+
+export const fetchWork = (workId) => {
+  const recordUrl = apiUrl + recordPath;
+  return (dispatch) => {
+    return axios.get(recordUrl, { params: { recordID: workId } })
+      .then((resp) => {
+        if (resp.data) {
+          dispatch(workDetail(resp.data));
+          // Do I need to change state of the searchResults here?
```
EdwinGuzman: Hmm why do you need to change the state for `searchResults` if you are fetching on specific work? To answer the question, no, you would update the state in the reducer and not in the action.

src/app/components/SearchResults/SearchResults.jsx
```diff
@@ -14,7 +14,7 @@ const SearchResults = (props) => {
     <div className="nypl-row">
       <div className="nypl-column-full">
         <ResultsMetadata metadata={metadata} />
-        <ResultsList results={hits} />
+        <ResultsList results={hits} actions={props} />
```
EdwinGuzman: I would pass in `fetchWork={props.fetchWork}` instead of `actions={props}` because props includes `results` and other data you may not need or have already pass in (like `hits`). Then in `ResultsList` you can call `props.fetchWork(workId)`.

src/app/components/WorkDetail/WorkDetail.jsx
EdwinGuzman: This doesn't work and it'll be easier to make this work if the search component routes to `/?author=....` itself.

src/app/components/WorkDetail/WorkDetail.jsx
```diff
@@ -0,0 +1,110 @@
...
+            <Link to={{ pathname: `${ebookUrl}`, query: { url: `${ebook.url}` } }}>{ebook.epub_path.split('/').pop()}</Link><br />
+          </span>
+        ))}
+      </div>
+    );
+  };
+
+  const parseInstances = (instances) => {
+    return (
+      <table className="nypl-basic-table">
```
EdwinGuzman: This is missing some accessibility attributes but I'd say make that a ticket for later.

src/app/components/WorkDetail/WorkDetail.jsx
```diff
@@ -0,0 +1,110 @@
...
+            <dd>
+              {item.title}
+            </dd>
+            <dt>Authors, Creators, et al</dt>
+            <dd>
+              {parseEntities(item.entities)}
+            </dd>
+            <dt>Subjects</dt>
+            <dd className="subject-listing">
+              {parseSubjects(item.subjects)}
```
EdwinGuzman: What happens if you don't have any subjects? Then you'll have an empty `Subjects` heading. You can do something like
```
{ item.subjects.length &&
  (
    <dt>Subjects</dt>
   <dd>....</dd>
  )
}
```

(This won't work because you can't render two siblings like this, but this is the general idea).

src/app/routes/routes.jsx
```diff
@@ -8,6 +8,7 @@ const routes = {
   default: (
     <Route path="/" component={Application}>
       <Route path="/search" component={Application} />
+      <Route path="/work" component={Application} />
```
EdwinGuzman: When you click on a work's title, you do not go to that route even though the component loads. What you want to do is update the component here to be `WorkDetail` to render that. Also, the `/search` path should load the `SearchResults` component. This would be tied down together with makes changes in the `Application` component so that you can render the children... something like
```
   <Application>
     <SearchContainer>
       <SearchForm>  //I am assuming you always want to render this component
         // either render
         <SearchResults> // for the /search path
         <WorkDetail> // for the /work path
```

Let's go over that next week? For now and for this PR, the rendering of different components is working and fine.

src/server/ApiRoutes/Search.js
```diff
@@ -1,10 +1,19 @@
...
+  const { q } = getRequestParams(req.query);
 
-  res.data = search(q, filters);
+  res.data = searchGet(q);
+
+  next();
+};
+
+export const getWorkById = (req, res, next) => {
+  const { workId } = req.query.workId;
```
EdwinGuzman: For this to work you want to destructure the object where the value is stored. So `const { workId } = req.query;` This is useful of `req.query` is an object with multiple properties and you only want to extract a few. If you only need `workId` you can do `const workId = req.query.workId` without the braces.

src/app/components/SearchForm/SearchForm.jsx
```diff
@@ -34,9 +34,17 @@ class SearchForm extends React.Component {
...
+      throw new Error('Please enter a term or terms to search');
+    }
 
     const terms = this.state.searchQuery.trim().replace(/\s+/g, ' ');
-    this.props.search(terms, this.state.searchField);
+
+    if (!this.state.searchField || this.state.searchField === 'q') {
+      this.props.searchGet(terms);
+    } else {
+      this.props.searchPost(terms, this.state.searchField);
```
EdwinGuzman: Why the two different actions?

EdwinGuzman:
There are some changes that you should make before approving and merging, but there are also larger architectural changes that we can go over together next week.

src/app/components/SearchContainer/SearchContainer.jsx
gkallenberg: Removing

src/app/components/WorkDetail/WorkDetail.jsx
gkallenberg: I guess it depends on how role is being handled in the data model. Is it a string or an array or either? That may dictate how the output is parsed.

src/app/actions/SearchActions.js
gkallenberg: This shows my lack of React knowledge and understanding since I wasn't able to just show the details without also showing the results. It's unclear to me how "pages" are handled in React where one display only shows one set of props from state while hiding other props from state. Perhaps I need to switch containers?

src/app/components/WorkDetail/WorkDetail.jsx
gkallenberg: Right -- I need a better way to parse the sections of the details page. It may be possible to just include the `dt` and `dd` elements in the same parsing function.

src/app/routes/routes.jsx
gkallenberg: Agreed that this needs to change in the routing. Am also having difficulty with server-side rendering with the Redux store and associated actions. Let's discuss next week for sure. 😄

src/server/ApiRoutes/Search.js
gkallenberg: Deconstructing properly would be better -- I'll adjust.

src/app/components/SearchForm/SearchForm.jsx
gkallenberg: At present, we're using searchGet method for general keyword searching across available fields and the searchPost method to handle field specific and advanced search queries which specify one or more fields with filtering and aggregations. The search API is fairly simple for now and doesn't do more sophisticated parsing like the Discovery API might be doing with its more complicate GET requests.

PR #7, opened 2019-01-25
Server route prep

gkallenberg:
This PR adds integration with React Router in order to render children at the Application level and provide location changes for navigating to different routes in the browser. A Breadcrumbs component was added to do some preliminary testing of parsing routes for navigation links on pages as part of the router integration. This can be removed since it's not vital to the operation of the site at this point.

The WorkDetails page now uses a DefinitionList component to parse an item's elements and return the HTML for display. Am leaving the ordering of elements as they come from Elasticsearch for now since no decisions have been made as far as placement, actual style, etc.

Also added some server-side async methods to test their ability to retrieve and ultimately update the Redux store for the client-side.

This PR also fixes an issue where making a second search request from the search form shows no results. The application component test was commented out due to errors with rendering child components from the test environment.


src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -0,0 +1,130 @@
...
+   * Generate an unordered list of links to ebooks (EPUBs) to our reader for locally
+   * stored EPUBs or a download link otherwise.
+   * @param {object} ebooks
+   */
+  const parseEbooks = (ebooks) => {
+    return (
+      <ul className="nypl-items-list">
+        {ebooks.map((ebook, i) => (
+            <li key={i.toString()}>
+            <a href={`${ebookUrl}?url=${ebook.url}`}>Read Online</a> [{ebook.epub_path.split('/').pop()}]<br />
```
mwbenowitz: I like this solution!

src/app/actions/SearchActions.js
```diff
@@ -77,4 +76,48 @@ export const fetchWork = (workId) => {
...
+    })
+    .catch((error) => {
+      console.log('An error occurred during searchGet', error);
+      throw new Error('An error occurred during searchGet', error);
+    });
+};
+
+export const serverPost = (query, field) => {
+  // Need a parsed query input to use for each filter
+  const userQuery = (query) ? encodeURIComponent(query) : '*';
```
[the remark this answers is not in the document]
gkallenberg: I'm leaning toward POST only for queries read from the form POST and from URL query strings to cover the server-side rendering. We still need to sanitize any user input.

src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -23,20 +23,29 @@ class SearchContainer extends React.Component {
...
-        {!this.showingDetails &&
-        <SearchResults
-          results={this.props.searchResults}
-          {...this.boundActions}
-        />}
-        <WorkDetail detail={this.props.workDetail} />
-      </div>
+      <main id="mainContent">
+        <div className="nypl-page-header">
+          <div className="breadcrumb" />
```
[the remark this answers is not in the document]
gkallenberg: Placeholder for the nav elements for breadcrumbs. This is partially done recycling code form discovery-front-end, but it doesn't quite work yet since I need to figure out handling route history and server-side rendering -- as well as caching/checking store before rendering to avoid unnecessary calls to the API.

src/app/components/WorkDetail/DefinitionList.jsx
gkallenberg: For SFR-72, I'm going to add a new, reusable component to handle this logic for the results list and details pages.

server.js
```diff
@@ -14,6 +14,7 @@ import routes from './src/app/routes/routes';
 import store from './src/app/stores/ReduxStore';
 import appConfig from './appConfig';
 import webpackConfig from './webpack.config';
+import { searchServer } from './src/server/ApiRoutes/Search';
```
EdwinGuzman: This is not used in this file.

src/app/actions/SearchActions.js
```diff
@@ -77,4 +76,48 @@ export const fetchWork = (workId) => {
...
 };
 
-export default { searchGet, searchPost, fetchWork };
+export const serverGet = (query) => {
+  // Need a parsed query input to use for each filter
+  const userQuery = (query) ? encodeURIComponent(query) : '*';
+  if (!userQuery) {
+    throw new Error('No search terms were entered. Please enter some terms.');
+  }
+  axios.get(searchUrl, { params: { q: userQuery } })
```
EdwinGuzman: These functions work without you having to explicitly `return` them?

src/app/components/Application/Application.jsx
```diff
@@ -1,10 +1,17 @@
...
 
 import { Header, navConfig } from '@nypl/dgx-header-component';
 import Footer from '@nypl/dgx-react-footer';
-import SearchContainer from '../SearchContainer/SearchContainer';
 
 class Application extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.props = props;
```
EdwinGuzman: You shouldn't need to do this. The `props` in `this` are already what's coming in from the constructor.

src/app/routes/routes.jsx
```diff
@@ -1,14 +1,18 @@
...
 import NotFound404 from '../components/Error/NotFound404';
+import WorkDetail from '../components/WorkDetail/WorkDetail';
+import Home from '../components/Home/Home';
 
 const routes = {
   default: (
     <Route path="/" component={Application}>
-      <Route path="/search" component={Application} />
-      <Route path="/work" component={Application} />
+      <IndexRoute component={Home} />
```
EdwinGuzman: This doesn't work if it's `SearchContainer` instead of `Home`?

gkallenberg: I'll try with `SearchContainer` to see if that satisfies React Router.

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -0,0 +1,130 @@
...
+      </ul>
+    );
+  };
+
+  /**
+   * Wrapper function to handle building the HTML from the object given.
+   * @param {object} details
+   */
+  const getDefinitions = (details) => {
+    const detailsObject = details.data.map(entry => (
```
EdwinGuzman: There's no check to see if `details.data`  (or `data` at the higher prop level) is not empty.

test/unit/Application.test.js
```diff
@@ -1,25 +1,25 @@
-/* eslint-env mocha */
-import React from 'react';
-import { expect } from 'chai';
-import { shallow } from 'enzyme';
-import Application from '../../src/app/components/Application/Application';
-import appConfig from '../../appConfig';
+// /* eslint-env mocha */
```
EdwinGuzman: If you don't need these just remove instead of comment out?

test/unit/DefinitionList.test.js
```diff
@@ -0,0 +1,48 @@
+/* eslint-env mocha */
+import React from 'react';
+import { expect } from 'chai';
+import { shallow } from 'enzyme';
+import { DefinitionList, labels } from '../../src/app/components/WorkDetail/DefinitionList';
+import { detail } from '../fixtures/work-detail.json';
```
EdwinGuzman: 👍

PR #10, opened 2019-02-12
Server side rendering

gkallenberg:
PR adds server side rendering which allows us to feed information to search engines and power the site with JS turned off. Additional enhancements made to help streamline code and more testing.

* adds a configureable store for adding more than one middleware package
* fixes server axios calls to send a properly configured state object
* adds tests for the SearchContainer and more for SearchForm components
* removes the GET methods since we're defaulting to POST as the standard method
* better default variable handling for query parameters
* uses constants.js for translating standard field names to ES field names
* adds comments to some methods
* prep work on breadcrumbs component


src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -7,37 +7,45 @@ import SearchForm from '../SearchForm/SearchForm';
 import SearchResults from '../SearchResults/SearchResults';
 import * as searchActions from '../../actions/SearchActions';
 
+/**
+ * Container class providing the Redux action creators
+ * to its child components. State data is passed along
+ * including params set in location.
+ *
+ * Accessibility Note: Creates the <main> element for all
+ * search pages with the corresponding <h1>.
```
EdwinGuzman: I think only the homepage will need the `h1`.

src/app/search/query.js
```diff
@@ -1,7 +1,7 @@
 export const getRequestParams = (query = {}) => {
-  const { q } = query || '*';
-  const { field } = query || 'title';
-  const { workId } = query || '';
+  const { q = '*' } = query;
```
EdwinGuzman: Nice 👌

EdwinGuzman:
It looks good. I get this error but not consistently (usually when making a new search and clicking on a title):
screen shot 2019-02-13 at 3 21 38 pm

gkallenberg:
The error you point out seems to happen when using server-side fetching say directly adding a search URL like: /search?q=bronte -- when clicking the title, the /work?workId=[uuid] doesn't seem to populate the item detail prop for the WorkDetail component on the first click. The second click on the title will load the WorkDetail page properly.

I need to figure out why the details prop doesn't get populated initially to load the WorkDetail page.

EdwinGuzman:
I don't think it's the server side fetching but the client side call to the work page. I'm not sure why it specifically doesn't work on the first hit and then works for subsequent hits.

In `ResultsList.jsx` and in the `showWorkDetail()` call, the function makes an async call to fetch the work but doesn't wait until the data is ready and instead it goes to the /work page. If you do something like:
```
this.props.fetchWork(workId)
  .then(() => {
    this.context.router.push(`/work?workId=${workId}`);
});
```
it should wait until the data is fetched and then go to the page. You can also add a catch to prevent it from going to the /work page if there is no data.

gkallenberg:
Right -- adding this now to both calls to the async functions (SearchForm and WorkDetail). Will test again to see if the server side is triggered when using the `to` parameter on the `<Link>` component.

EdwinGuzman:
Works well now 👌

src/app/components/SearchForm/SearchForm.jsx
```diff
@@ -32,27 +32,25 @@ class SearchForm extends React.Component {
...
 
-    if (!this.state.searchField || this.state.searchField === 'q') {
-      this.props.searchGet(terms);
-    } else {
-      this.props.searchPost(terms, this.state.searchField);
-    }
-    this.context.router.push(`/search?q=${terms}`);
+    this.props.searchPost(terms, this.state.searchField)
+      .then(() => {
+        this.context.router.push(`/search?q=${terms}`);
```
mwbenowitz: Would it make sense to show the searchField here instead of just "q"?

EdwinGuzman:
Yeah! It seems the new version of node doesn't really like older versions of sass. You can do something like `npm rebuild --node-sass`. We use `3.8.0` for everything so it's worth looking into updating it for all the other apps, _eventually_ 😞.

mwbenowitz:
ah ok! I always seem to run into the fun version conflicts.

src/app/components/SearchForm/SearchForm.jsx
gkallenberg: Am exposing the field as a second parameter in the next PR for the Title, Author and to be added Subject fields since we'll need to pass those along. More URL params to come when adding sort, count per page, page, etc.

PR #11, opened 2019-02-15
Title (field) search

gkallenberg:
Adds `field` to the location when searching via the client. The server should also respond to the `field` query parameter. Query objects are now built using a utility function which I'm hoping I can expand to adding filter, aggregations, sort, page and per_page elements. Some tests have been added just for testing the field queries built to send along to ES.

In the query builder, phrases are split into words, encoding moved here, and each word is added as separate field queries in an array which acts as an AND for all words so each must be present in the field. The query builder removes some punctuation including commas to split words properly which helps for author searches. This PR should satisfy all basic field level querying which need to have terms AND'd together.

### DIFF
--- a/src/app/search/query.js
+++ b/src/app/search/query.js
@@ -22,12 +22,17 @@ const addFieldQuery = (queryString, field = 'keyword') => {
   }
   let fieldQuery = [];
   // Strip punctuation and process spaces as plus signs for final split.
-  const queryArr = queryString.replace(/[.,\/#!$%\^&;:{}=\-_`~()]/g, '').trim().replace(/\s+/g, '+').split('+');
+  const queryArr = queryString.replace(/[.\/#!$%\^&;{}=\-_`~()]/g, '').trim().replace(/\s+/g, '+').split('+');
+  /**
+   * Reformat the string.
+   * For multiple strings in a query, add the plus operator to AND terms together;
+   * otherwise, return a single string.
+   */
+  const esQuery = (queryArr.length > 1) ? queryArr.join(' +') : queryArr.join('');
 
-  // For each string in queryArr, add a field query object.
-  queryArr.map(query => (
-    fieldQuery.push({ field, value: encodeURIComponent(query) })
-  ));
+  // TODO: add an additional check on empty queries after the terms are processed.
+
+  fieldQuery.push({ field, value: encodeURIComponent(esQuery) });
 
   return fieldQuery;
 };
--- a/src/app/search/query.js
+++ b/src/app/search/query.js
@@ -1,3 +1,4 @@
+
 export const getRequestParams = (query = {}) => {
   const { q = '*' } = query;
   const { field = 'keyword' } = query;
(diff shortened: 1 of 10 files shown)


src/app/search/query.js
```diff
@@ -6,4 +6,71 @@ export const getRequestParams = (query = {}) => {
   return { q, field, workId };
 };
 
-export default getRequestParams;
+/**
+ * Format a single field query or field queries based on string sent. Strings
+ * are broken up by word to AND them together by adding a separate field/value
```
EdwinGuzman: This sentence is confusing.

test/unit/SearchActions.test.js
```diff
@@ -1,25 +1,43 @@
...
+          {
+            title: 'The Tragedie of Hamlet, Prince of Denmark\nA Study with the Text of the Folio of 1623',
+            rights_stmt: 'Public domain in the USA.',
+          },
+        },
+      };
+      workDetailAction = workDetail(details);
+
+      expect(workDetailAction.type).to.equal('FETCH_WORK');
+      expect(JSON.stringify(workDetailAction.item)).to.equal(JSON.stringify(details));
```
EdwinGuzman: `workDetailAction.item.detail.item` ?!

test/unit/Query.test.js
```diff
@@ -0,0 +1,38 @@
...
+    expect(parsedValueOne.indexOf('$')).to.equal(-1);
+    expect(parsedValueOne.indexOf(',')).to.equal(-1);
+    expect(parsedValueOne.indexOf('\'')).to.not.equal(-1);
+    expect(parsedValueTwo.indexOf(';')).to.equal(-1);
+    expect(parsedValueFour).to.equal('*');
+  });
+
+  it('should return field query objects for each word within the "queries" array.', () => {
+    const queryBody = buildQueryBody(query);
+    expect(queryBody.queries.length).to.equal(4);
```
EdwinGuzman: You should also test each word in the queries array is the word you are expecting in order.

src/app/search/query.js
```diff
@@ -6,4 +6,71 @@ export const getRequestParams = (query = {}) => {
...
+ *
+ * @returns {string}
+ */
+const addFieldQuery = (queryString, field = 'keyword') => {
+  if (!queryString) {
+    throw new Error('A valid query string must be passed');
+  }
+  let fieldQuery = [];
+  // Strip punctuation and process spaces as plus signs for final split.
+  const queryArr = queryString.replace(/[.,\/#!$%\^&;:{}=\-_`~()]/g, '').trim().replace(/\s+/g, '+').split('+');
```
mwbenowitz: I agree with stripping these, but I can see some characters especially commas and colons being used in valid searches?

Also, an extreme edge case, but searching for _just_ punctuation returns nothing but it appears in the URL. Should probably throw the same error as an empty search string?

src/app/search/query.js
gkallenberg: Fixing this to be a bit less vague.

test/unit/SearchActions.test.js
gkallenberg: am refactoring the object so it's not quite so nested or nested with repeating labels.

src/app/search/query.js
gkallenberg: As far as the first question, I've decided to alter the field query build a bit by sending a query like "hell +poetry" multiple terms instead of multiple field objects. I think that will work better.

I've added a TODO in comments for the latter. I need another PR, I think to manage errors and display them back to the user. Strange queries that become empty after processing should just return an empty result set for now.

EdwinGuzman:
Last commit looks good 👍

mwbenowitz:
Agree, looks good!

PR #12, opened 2019-02-20
Add subject search

gkallenberg:
* adds subject search option to form
* consolidates option value rendering on SearchForm 
* adds the change-case npm package for capitalizing labels for search form options and other potential uses in the project. If there's a better one, just let me know.
* adds onClick handler for author and subject links on detail pages to the DefinitionList component
** passed dispatch and context from WorkDetail component for onClick handler

### DIFF
--- a/src/app/components/SearchForm/SearchForm.jsx
+++ b/src/app/components/SearchForm/SearchForm.jsx
@@ -63,17 +63,10 @@ class SearchForm extends React.Component {
                       onChange={this.onFieldChange}
                       value={this.state.searchField}
                     >
-<<<<<<< HEAD
                       {this.props.allowedFields.map((field, key) => {
                         return (
                           <option value={field} key={key.toString()}>
                             {titleCase(field)}
-=======
-                      {this.props.allowedFields.map((element, key) => {
-                        return (
-                          <option value={element} key={key.toString()}>
-                            {titleCase(element)}
->>>>>>> bc08aa6e8290752d4ad3e41e002d1dca9a7f0847
                           </option>
                         );
                       })}
@@ -89,11 +82,7 @@ class SearchForm extends React.Component {
                       type="text"
                       aria-labelledby="search-input-field"
                       value={this.state.searchQuery}
-<<<<<<< HEAD
                       placeholder="Keyword, title, author, or subject"
-=======
-                      placeholder="Keyword, title, author or subject"
->>>>>>> bc08aa6e8290752d4ad3e41e002d1dca9a7f0847
                       onChange={this.onQueryChange}
                     />
                   </span>
--- a/src/app/components/SearchForm/SearchForm.jsx
+++ b/src/app/components/SearchForm/SearchForm.jsx
@@ -43,7 +43,8 @@ class SearchForm extends React.Component {
 
     this.props.searchPost(terms, this.state.searchField)
       .then(() => {
-        this.context.router.push(`/search?q=${terms}&field=${this.state.searchField}`);
+        const encodedUserInput = encodeURIComponent(terms);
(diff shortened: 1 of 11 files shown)


src/app/components/SearchForm/SearchForm.jsx
```diff
@@ -61,15 +62,13 @@ class SearchForm extends React.Component {
...
-                      <option value={this.props.allowedFields.kw} defaultValue>
-                        Keyword
-                      </option>
-                      <option value={this.props.allowedFields.ti}>
-                        Title
-                      </option>
-                      <option value={this.props.allowedFields.au}>
-                        Author
-                      </option>
+                      {this.props.allowedFields.map((element, key) => {
```
EdwinGuzman: I might change the `element` variable name to something different since you can pass HTML elements around in jsx and that could be confusing.

src/app/components/SearchForm/SearchForm.jsx
```diff
@@ -82,7 +81,7 @@ class SearchForm extends React.Component {
                       type="text"
                       aria-labelledby="search-input-field"
                       value={this.state.searchQuery}
-                      placeholder="Keyword, title, or author"
+                      placeholder="Keyword, title, author or subject"
```
EdwinGuzman: `author, or ` (comma after author)?

EdwinGuzman:
* When I search for `war` the first result is `The soul of the war`. I then go to that work and click on its subject `World War, 1914-1918 -- Personal narratives`. That takes me to a search results page but I was expecting to see `The soul of the war` but it is not there. Don't think that's necessarily a front-end issue but just pointing it out.
* When I build the project `npm run dist` and then run the app in production mode `npm run prod-start`,  I get this warning in an incognito window which seems related to this PR:

screen shot 2019-02-21 at 10 41 03 am

src/app/components/SearchForm/SearchForm.jsx
gkallenberg: Fixing to use `field` instead of `element`

src/app/components/SearchForm/SearchForm.jsx
gkallenberg: Comma incoming.

gkallenberg:
Queries are not handled properly and are being refactored slightly to improve phrase searching and coincide with slight changes to the search API which Mike has employed. The changes should be a step towards better processing of user inputs before sending to the API to query Elasticsearch. Although, changes may not be ideal, they should move us in the right direction for query string searching.

For the connections errors, I may need to add the development address to the production variable in appConfig until we have a production box in place for the index and API.

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -17,13 +18,31 @@ export const labels = {
...
+   * onClick handler for browse searches.
+   *
+   * @param {object} event
+   * @param {string} query
+   * @param {string} field
+   */
+  const newSearchRequest = (event, query, field) => {
+    event.preventDefault();
+
+    props.dispatch(searchPost(query, field))
```
mwbenowitz: In testing this doesn't seem to make a difference, but we might want to wrap the `query` here in quotes to force the API/ElasticSearch to only return exact matches. Generally this is what the `AND`ed queries will do, but for a simple subject with some common terms it _could_ have unintended results.

gkallenberg: Not sure if we can add quotes around the query there without another parameter signaling that the query is a "browse" query coming from a link rather than an ad-hoc query typed in the search box or URL. But seems like a decision for another PR to handle.

PR #15, opened 2019-03-15
Add Breadcrumbs

gkallenberg:
The breadcrumbs component adds links in a trail from the home page of ResearchNow to the details view. One should be able to navigate back to either the search results one just came from if on a detail page, or back to the home page with a fresh start. Breadcrumbs should work the same whether or not the search was initiated from the form, a browse link, or a request via the address bar, e.g. /search?q=journey&field=title. Includes tests and some inline documentation.

### DIFF
--- a/src/app/components/Breadcrumbs/Breadcrumbs.jsx
+++ b/src/app/components/Breadcrumbs/Breadcrumbs.jsx
@@ -25,17 +25,13 @@ const Breadcrumbs = (({ links, pageType, onClickHandler }) => {
   const crumbTrail = () => {
     const crumbs = [homeLink];
 
-    if (links && Array.isArray(links)) {
-      links.map((link, iterator) => {
-        if (iterator < links.length-1) {
-          crumbs.push(
-            <li key={iterator}>
-              <Link to={link.href}>
-                {link.text}
-              </Link>
-            </li>);
+    if (links && links.length) {
+      links.forEach((link, iterator) => {
+        const linkKey = `links-${iterator}`;
+        if (iterator < links.length - 1) {
+          crumbs.push(<li key={linkKey}><Link href={link.href}>{link.text}</Link></li>);
         } else {
-          crumbs.push(<li key={iterator}>{link.text}</li>);
+          crumbs.push(<li key={linkKey}>{link.text}</li>);
         }
       });
     }
--- a/src/app/components/Breadcrumbs/Breadcrumbs.jsx
+++ b/src/app/components/Breadcrumbs/Breadcrumbs.jsx
@@ -8,7 +8,7 @@ import { Link } from 'react-router';
  * @param {object} props
  * @returns {array}
  */
-const Breadcrumbs = (({ query = '', type, workUrl }) => {
+const Breadcrumbs = (({ query = '', field = '', type }) => {
   if (type === 'home') {
     return null;
   }
@@ -17,7 +17,7 @@ const Breadcrumbs = (({ query = '', type, workUrl }) => {
 
   const homeLink = (
     <li key="home">
-      <Link to={'/'} onClick={() => onClick('ResarchNow')}>
+      <Link to={'/'}>
         ResearchNow
       </Link>
     </li>);
@@ -32,7 +32,7 @@ const Breadcrumbs = (({ query = '', type, workUrl }) => {
 
     crumbs.push(
       <li key="results">
-        <Link to={`/search?q=${query}`} onClick={() => onClick('Search Results')}>
+        <Link to={`/search?q=${query}&field=${field}`}>
           Search Results
         </Link>
       </li>);
@@ -44,7 +44,7 @@ const Breadcrumbs = (({ query = '', type, workUrl }) => {
 
     crumbs.push(
       <li key="details">
-        <Link to={`/work`} onClick={() => onClick('Work Details')}>Work Details</Link>
+        <Link to={`/work`}>Work Details</Link>
       </li>);
 
     return crumbs;
@@ -64,6 +64,7 @@ const Breadcrumbs = (({ query = '', type, workUrl }) => {
 
 Breadcrumbs.propTypes = {
   query: PropTypes.string,
+  field: PropTypes.string,
   type: PropTypes.string,
 };
 
--- a/src/app/components/Breadcrumbs/Breadcrumbs.jsx
+++ b/src/app/components/Breadcrumbs/Breadcrumbs.jsx
@@ -29,7 +29,7 @@ const Breadcrumbs = (({ links, pageType, onClickHandler }) => {
       links.forEach((link, iterator) => {
         const linkKey = `links-${iterator}`;
         if (iterator < links.length - 1) {
-          crumbs.push(<li key={linkKey}><Link href={link.href}>{link.text}</Link></li>);
+          crumbs.push(<li key={linkKey}><Link to={link.href}>{link.text}</Link></li>);
         } else {
           crumbs.push(<li key={linkKey}>{link.text}</li>);
         }
--- a/src/app/components/Breadcrumbs/Breadcrumbs.jsx
+++ b/src/app/components/Breadcrumbs/Breadcrumbs.jsx
@@ -8,8 +8,8 @@ import { Link } from 'react-router';
  * @param {object} props
  * @returns {array}
  */
-const Breadcrumbs = (({ query = '', field = '', type }) => {
-  if (type === 'home') {
+const Breadcrumbs = (({ query = '', field = '', pageType, handleReset }) => {
+  if (pageType === 'home') {
     return null;
   }
 
@@ -17,7 +17,7 @@ const Breadcrumbs = (({ query = '', field = '', type }) => {
 
   const homeLink = (
     <li key="home">
-      <Link to={'/'}>
+      <Link to={'/'} onClick={event => handleReset(event)}>
         ResearchNow
       </Link>
     </li>);
@@ -25,7 +25,7 @@ const Breadcrumbs = (({ query = '', field = '', type }) => {
   const crumbTrail = () => {
     const crumbs = [homeLink];
 
-    if (type === 'results') {
+    if (pageType === 'results') {
       crumbs.push(<li key="results">Search Results</li>);
       return crumbs;
     }
@@ -37,7 +37,7 @@ const Breadcrumbs = (({ query = '', field = '', type }) => {
         </Link>
       </li>);
 
-    if (type === 'details') {
+    if (pageType === 'details') {
       crumbs.push(<li key="details">Work Detail</li>);
       return crumbs;
     }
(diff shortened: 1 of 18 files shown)


src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -23,36 +26,45 @@ class SearchContainer extends React.Component {
...
-            <div className="nypl-column-full">
-              <h1 className="nypl-heading">ResearchNow</h1>
-              <div id="tagline">
-                Search the world&apos;s research collections and more for digital books you
-                can use right now.
+        <div className="nypl-full-width-wrapper">
+          <div className="nypl-page-header">
+            <Breadcrumbs query={this.props.searchQuery} pageType={pageType} />
+          </div>
+          <div role="search" aria-label="ResearchNow">
```
EdwinGuzman: I think you want to use `role="search"` with the form https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Search_role.

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -32,6 +32,8 @@ export const DefinitionList = (props) => {
   const newSearchRequest = (event, query, field) => {
     event.preventDefault();
 
+    props.dispatch(userQuery(query));
+    props.dispatch(selectedField(field));
```
EdwinGuzman: I don't know what the best practice around this is, whether to just trigger one action per event or multiple actions per event.

EdwinGuzman:
This works well. I am getting and issue where it seems that a search term is "stuck" in the state of the app so even when I open a new incognito window and go to a _new_ search result item's page and click on the "Search Results" link, I still get the previous search term.

So if I search for "shakespeare" and then click on an item and share it, but then search for "cat" and share it, "shakespeare" will still be in the state.

Not an issue for this PR but if you are on the same search results page and keep searching new terms, when you hit the back button the browser the search term nor the results update accordingly (but the url updates correctly). The main part of this PR, the breadcrumbs, do work as they are suppose to.

The breacrumbs component is mostly based on what I did for Discovery. Do you think the logic for that is good? I wanted to make it more generic so you can pass in a list of the pages that should be rendered/linked but had to be content with hard coding values.

src/app/components/WorkDetail/WorkDetail.jsx
```diff
@@ -53,23 +78,30 @@ class WorkDetail extends React.Component {
 
 WorkDetail.propTypes = {
   work: PropTypes.object,
```
mwbenowitz: My esLint is getting mad at this line saying that `PropTypes.object` is disabled by default. I think it only raises a warning

src/app/components/SearchContainer/SearchContainer.jsx
gkallenberg: We should add role to the form as per best practice. Am asking for clarification from UX and Accessibility whether this landmark is appropriately placed or if it should be adjusted. Will wait for their answer and add as a separate PR.

src/app/components/WorkDetail/DefinitionList.jsx
gkallenberg: It's probably better to add a reset action to change state to empty values where appropriate. Am not sure how that will work with Redux, but will add a new action/reducer to handle a reset instead of calling three actions.

src/app/components/WorkDetail/WorkDetail.jsx
gkallenberg: Yes, it objects to the object designation in multiple places.

mwbenowitz:
I did some testing on this and I get the same error as you mentioned in stand up. What it seems to be is that the reducer doesn't get triggered. The action is recorded, but it never invokes resetSearch. I can't figure out why though.

src/app/components/SearchContainer/SearchContainer.jsx
EdwinGuzman: What was the consensus from the UX team?

src/app/components/WorkDetail/DefinitionList.jsx
EdwinGuzman: Yes, a reset action/reducer has worked for me in the past. I think we've called the reset action _after_ the initial action has finished.

src/app/components/SearchForm/SearchForm.jsx
```diff
@@ -15,6 +16,21 @@ class SearchForm extends React.Component {
     this.submitSearchRequest = this.submitSearchRequest.bind(this);
   }
 
+  /**
+   * Used to update the downstream props updated by the
+   * parent component, SearchContainer.
+   *
+   * @param {object} nextProps
+   */
+  componentWillReceiveProps(nextProps) {
```
EdwinGuzman: This component could also benefit from a `window.scrollTo()` call. When I am on a Work Detail page and I click on a subject or author link, the new search works well but the page doesn't go to the top of the results. _Maybe_ it would be better if the first result is focused or maybe a text descriptor explaining the new set of results (e.g. the "Displaying 1 - 2 ....").

EdwinGuzman:
I have not encountered the second to last issue mentioned about clicking on the `ResearchNow` breadcrumb link from a `Work Details` page and the search results not clearing out. I can get to the homepage consistently with blank search form. What's the other issue?

src/app/components/SearchForm/SearchForm.jsx
gkallenberg: The set of components that handle messaging, sorting, paging still need work which is why the message is so generic for now. Definitely wasn't sure the best way to handling scrolling either:  what method is best, what components need it, etc.

src/app/components/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -2,37 +2,43 @@ import React from 'react';
...
-    crumbs.push(
-      <li key="results">
-        <Link to={`/search?q=${query.q}`} onClick={() => onClick('Search Results')}>
-          Search Results
-        </Link>
-      </li>);
-
-    if (type === 'details') {
-      crumbs.push(<li key="details">Work Details</li>);
+    if (links && Array.isArray(links)) {
```
EdwinGuzman: `links.length` may be a better check to make sure its not empty.

src/app/components/SearchForm/SearchForm.jsx
EdwinGuzman: The message being generic is okay. The UX team can handle where to focus to in that case to see what method is best.

src/app/components/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -2,37 +2,43 @@ import React from 'react';
...
+      links.map((link, iterator) => {
```
nachocs: Probably better a forEach here, as the value returned by map is not used

src/app/stores/Reducers.js
```diff
@@ -23,35 +24,50 @@ export const workDetail = (state = null, action) => {
...
-  field,
-  allowedFilters,
+  searchQuery,
+  searchField,
   sort,
   workDetail,
 });
+
+export const rootReducer = (state, action) => {
+  if (action.type === Actions.RESET_SEARCH) {
```
nachocs: wouldn't be better to have Reset_Search with it's own reducer? - like the one commented out -

src/app/stores/Reducers.js
```diff
@@ -23,35 +24,50 @@ export const workDetail = (state = null, action) => {
...
+    state = initialState;
+  }
+
+  return appReducer(state, action);
+};
+
+export default rootReducer;
```
EdwinGuzman: Here you are exporting `rootReducer` twice, once as the default and once as a variable. Right now you are importing it as `import searchReducer from './Reducers';` which would be this `default` export. But you can also do `import { rootReducer } from './Reducers';` and the app would work just the same. For consistency and to reduce confusion, pick one or the other.

The default option may be the best one since you currently only do the former import.

src/app/stores/Reducers.js
gkallenberg: I used this answer given to this question by Dan Abramov as an example where the rootReducer resets the state object when a certain action is fired: https://stackoverflow.com/questions/35622588/how-to-reset-the-state-of-a-redux-store. For some reason, the Reducer used separately to return initialState didn't filter down, but I might not have had the correct lifecycle methods running on the components to understand there was a state change.

src/app/stores/Reducers.js
gkallenberg: Changing to reflect the reducer name change.

src/app/components/Breadcrumbs/Breadcrumbs.jsx
gkallenberg: Changing

src/app/components/Breadcrumbs/Breadcrumbs.jsx
gkallenberg: Changing over to forEach here.

src/app/components/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -2,37 +2,43 @@ import React from 'react';
...
+        if (iterator < links.length-1) {
+          crumbs.push(
+            <li key={iterator}>
+              <Link to={link.href}>
```
gkallenberg: Is it better to use the `href` property here instead of `to`?

EdwinGuzman: `to` is specific to `Link` so it's fine.

test/unit/Breadcrumbs.test.js
```diff
@@ -0,0 +1,101 @@
...
+    let component;
+    const links = [{ href: '', text: '' }];
+    const pageType = 'home';
+
+    before(() => {
+      component = shallow(<Breadcrumbs links={links} pageType={pageType} />);
+    });
+
+    it('should display no breadcrumb nav element', () => {
+      expect(component).to.be.unrendered;
```
EdwinGuzman: Nice. Didn't know there was an `unrendered`.

test/unit/Breadcrumbs.test.js
```diff
@@ -0,0 +1,101 @@
...
+      component = shallow(<Breadcrumbs links={links} pageType={pageType} />);
+    });
+
+    it('should render a nav and ol element', () => {
+      expect(component.find('nav').length).to.equal(1);
+      expect(component.find('ol').length).to.equal(1);
+    });
+
+    it('should have three list items and two links', () => {
+      expect(component.find('li').length).to.equal(3);
```
EdwinGuzman: Maybe being more explicit about what values these are will be better so you know that you have three li items but only two `Link`s. You have tests below for the `href` prop but having it will be nice too.

test/unit/Breadcrumbs.test.js
```diff
@@ -0,0 +1,101 @@
...
+    it('should display a link back to the results page and home page', () => {
+      const detailLinks = component.find('Link');
+      expect(detailLinks.length).to.equal(2);
+      expect(detailLinks.at(0).children().text()).to.equal('ResearchNow');
+      expect(detailLinks.at(0).prop('to')).to.equal('/');
+      expect(detailLinks.at(1).children().text()).to.equal('Search Results');
+      expect(detailLinks.at(1).prop('href')).to.equal('/search?q=journey&field=title');
+    });
+
+    it('should display a crumb for the details page without being a link', () => {
```
EdwinGuzman: If there's no `Link` then a `expect(component.find('Link').length).to.equal(0);` would be a good test to have.

EdwinGuzman:
Nice, just a few minor comments but looks good overall.

PR #17, opened 2019-03-28
Sfr 351 search by viaf

nachocs:
Changed searches API calls to work based upon their components, no need to call the api before and pages work by themselves (no need of server side) and in sync with the store.
Added search by viaf and lcnaf, adding a parameter to the search to show name of author in search box.
Added details about authors on work page and authors in search results.
Simplified some components.
Decode html entities in some values to avoid &amp in texts

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,8 @@
     "react-hot-loader": "1.2.8",
     "react-test-renderer": "^16.6.3",
     "redux-mock-store": "^1.5.3",
-    "sinon": "^4.1.3"
+    "sinon": "^4.1.3",
+    "webpack-dev-server": "2.9.7"
   },
   "dependencies": {
     "@nypl/design-toolkit": "0.1.36",
@@ -96,7 +97,6 @@
     "style-loader": "0.19.1",
     "underscore": "^1.9.1",
     "webpack": "2.7.0",
-    "webpack-dev-server": "2.9.7",
     "webpack-merge": "0.8.4"
   },
   "directories": {
--- a/package.json
+++ b/package.json
@@ -47,8 +47,7 @@
     "react-hot-loader": "1.2.8",
     "react-test-renderer": "^16.6.3",
     "redux-mock-store": "^1.5.3",
-    "sinon": "^4.1.3",
-    "webpack-dev-server": "2.9.7"
+    "sinon": "^4.1.3"
   },
   "dependencies": {
     "@nypl/design-toolkit": "0.1.36",
@@ -81,6 +80,7 @@
     "express": "4.16.2",
     "extract-text-webpack-plugin": "2.1.2",
     "history": "3.3.0",
+    "html-entities": "^1.2.1",
     "node-sass": "3.8.0",
     "prop-types": "15.6.2",
     "react": "15.5.4",
@@ -96,6 +96,7 @@
     "style-loader": "0.19.1",
     "underscore": "^1.9.1",
     "webpack": "2.7.0",
+    "webpack-dev-server": "2.9.7",
     "webpack-merge": "0.8.4"
   },
   "directories": {
--- a/server.js
+++ b/server.js
@@ -1,3 +1,5 @@
+/* eslint-disable global-require */
+/* eslint-disable react/jsx-filename-extension */
 import path from 'path';
 import express from 'express';
 import compress from 'compression';
@@ -55,11 +57,8 @@ app.get('/*', (req, res) => {
       }
       const store = configureStore(res.data);
 
-      const application = ReactDOMServer.renderToString(
-        <Provider store={store}>
-          <RouterContext {...renderProps} />
-        </Provider>
-      );
+      const application = ReactDOMServer
+        .renderToString(<Provider store={store}><RouterContext {...renderProps} /></Provider>);
 
       // First parameter references the ejs filename
       res.render('index', {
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -96,6 +96,7 @@
     "style-loader": "0.19.1",
     "underscore": "^1.9.1",
     "webpack": "2.7.0",
+    "webpack-dev-server": "2.9.7",
```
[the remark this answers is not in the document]
nachocs: I changed it because we were getting a warning to put it here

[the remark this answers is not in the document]
nachocs: changed back

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -21,28 +22,26 @@ export const labels = {
...
+      };
+    } else if (author.lcnaf) {
+      return {
+        q: author.lcnaf,
+        field: 'lcnaf',
+        showQuery: `\"${author.name}\"`,
+        showField: 'author',
+      };
+    }
+    return { q: `\"${author.name}\"`, field: 'author' };
```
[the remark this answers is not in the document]
nachocs: sure, changed it to more compact version

nachocs:
As we are changing the design, any test over that will have to change soon, so I think to leave them like this by the moment if it's ok

gkallenberg:
As far as testing, that's fine with me. We want to have proper test coverage which is part of our engineering best practices. We can circle back to these in my opinion.

server.js
```diff
@@ -55,11 +57,8 @@ app.get('/*', (req, res) => {
       }
       const store = configureStore(res.data);
 
-      const application = ReactDOMServer.renderToString(
-        <Provider store={store}>
-          <RouterContext {...renderProps} />
-        </Provider>
-      );
+      const application = ReactDOMServer
+        .renderToString(<Provider store={store}><RouterContext {...renderProps} /></Provider>);
```
EdwinGuzman: I prefer to have these on multiple lines to look more like formatted HTML but not a big deal.

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -21,28 +22,15 @@ export const labels = {
...
-
-    dispatch(userQuery(query));
-    dispatch(selectedField(field));
-    dispatch(searchPost(query, field))
-      .then(() => {
-        context.router.push(`/search?q=${query}&field=${field}`);
-      });
-  };
-
+  const getIdentifier = author => (author.viaf && 'viaf') || (author.lcnaf && 'lcnaf') || 'name';
```
EdwinGuzman: What is the `&& 'viaf'` check and the `&& 'lcnaf'` check for? Won't it always be true since non-empty strings always return true?

EdwinGuzman: Ah nevermind. I see what you are doing. Not necessarily the easier to read but nice.

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -81,11 +68,31 @@ export const DefinitionList = ({
...
             {list.map((entity, i) => (
               <li key={i.toString()}>
-                <Link
-                  onClick={event => newSearchRequest(event, `\"${entity.name}\"`, 'author')}
-                  to={{ pathname: '/search', query: { q: `\"${entity.name}\"`, field: 'author' } }}
-                >{entity.name}, {entity.roles.join(', ')}
+                <Link to={{ pathname: '/search', query: linkToAuthor(entity) }}>
+                  {htmlEntities.decode(entity.name)}, {entity.roles.join(', ')}
                 </Link>
+                {(entity.birth_date_display || entity.death_date_display) && <span> (</span>}
```
EdwinGuzman: This seems too complicated to leave like this and not in its own function.

EdwinGuzman:
I agree with Greg that there should be updated or new tests for PRs, in this case for the viaf and lncaf links.

I also got this error the first time I tried to click on a result from the search results:
Screen Shot 2019-03-29 at 4 12 00 PM

But then it went away. If you open up an incognito window and try it again, it'll come up every time for the first search. It seems the internal store keeps the value so the second time this doesn't happen.

src/app/components/WorkDetail/DefinitionList.jsx
nachocs: I'd prefer to use ternary, but it's not allowed in eslint rules.

nachocs:
I removed that throw Error, as it's ok to not have work on first load, let me know if you think it's ok

src/app/components/WorkDetail/DefinitionList.jsx
nachocs: changed

server.js
nachocs: the eslint rules made it like this when formating

EdwinGuzman:
Looks fine but returning and rendering `null` would be better than rendering an empty `div` which doesn't add anything.

nachocs:
Changed to return null

EdwinGuzman:
Looks good and I'll leave the final word to Greg. Tests may not be too important since they can change later when the design is updated, but having some base tests would be helpful, even if they change slightly.

gkallenberg:
Let's keep this moving forward and include tests for lcnaf and viaf links with the new design implementation. We're introducing the CMS.gov React components and USWDS CSS elements for the next PR.

PR #19, opened 2019-04-08
Sfr 345 udate design

nachocs:
Also SFR-374 Update design on Item


src/app/actions/SearchActions.js
```diff
@@ -60,25 +61,26 @@ export const searchPost = (query, field) => {
...
-    .then((resp) => {
-      if (resp.data) {
-        dispatch(workDetail(resp.data));
-      }
-    })
-    .catch((error) => {
-      console.log('An error occurred during fetchWork', error.message);
-      throw new Error('An error occurred during fetchWork', error.message);
-    });
+export const fetchWork = workId => dispatch => axios
```
EdwinGuzman: I think having `axios` on the next line makes it much more readable. This is very nice to see it is understandable but it's not easily readable. I'm not sure how you and Greg prefer but I prefer quick readability.

src/app/components/Application/Application.jsx
```diff
@@ -16,11 +16,7 @@ class Application extends React.Component {
   render() {
     return (
       <div className="app-wrapper">
-        <Header
-          skipNav={{ target: 'mainContent' }}
-          navData={navConfig.current}
-          urlType="absolute"
-        />
+        <Header skipNav={{ target: 'mainContent' }} navData={navConfig.current} urlType="absolute" />
```
EdwinGuzman: I think one of the linters also prefers having this as multiline.

src/app/components/Error/NotFound404.jsx
```diff
@@ -12,7 +12,10 @@ const NotFound404 = () => (
           <p>We&apos;re sorry...</p>
           <p>The page you were looking for doesn&apos;t exist.</p>
           <p>
-              Search <Link to={`${appConfig.baseUrl}/`}>ResearchNow</Link>.
+            Search
+            {' '}
```
EdwinGuzman: This seems unnecessary.

src/app/components/List/AuthorsList.jsx
```diff
@@ -0,0 +1,64 @@
...
+  showQuery: `"${author.name}"`,
+  showField: 'author',
+});
+const birthDate = entity => (
+  <span>
+    {(entity.birth_date_display || entity.death_date_display) && <span> (</span>}
+    {entity.birth_date_display && <span>{entity.birth_date_display}</span>}
+    {entity.death_date_display && (
+      <span>
+        {' '}
```
EdwinGuzman: Hmm I'm not liking these.

src/app/components/List/EditionsList.jsx
```diff
@@ -0,0 +1,90 @@
...
+    let publisher = instance && instance.agents && instance.agents.find(agent => agent.roles.indexOf('publisher') > -1);
+    publisher = publisher ? publisher.name : null;
+    return publisher;
+  };
+  const getIsValid = instance => (instance.items && instance.items.length > 0)
+    || (instance.pub_date && instance.pub_date_display)
+    || instance.pub_place
+    || getPublisher(instance);
+
+  const filterValid = instances => instances.filter(l => getIsValid(l));
```
EdwinGuzman: Why not use `instance` instead of `l` to be more explicit?

src/app/components/List/EditionsList.jsx
```diff
@@ -0,0 +1,90 @@
...
+        </tbody>
+      </table>
+      {!!max && filterValid(list).length > max && (
+        <div className="nypl-editions-view-all">
+          <Link to={{ pathname: '/work', query: { workId: `${work.uuid}` }, hash: '#all-editions' }}>
+            View All
+            {' '}
+            {filterValid(list).length}
+            {' '}
+Editions
```
EdwinGuzman: This is not indented well?

src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -30,39 +30,49 @@ class SearchContainer extends React.Component {
...
       global.window.scrollTo(0, 0);
       this.loadSearch();
     }
   }
 
   loadSearch() {
-    const { query } = this.props.location;
-    const searchQuery = query && query.q ? query.q : this.props.searchQuery;
-    const selectedField = query && query.field ? query.field : this.props.searchField;
+    let { searchQuery } = this.props;
```
EdwinGuzman: You should add `searchQuery` to the destructuring below and create a new variable for the update instead of overriding the prop value.

src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -30,39 +30,49 @@ class SearchContainer extends React.Component {
...
     }
   }
 
   render() {
-    const { query } = this.props.location;
+    const {
+      location: { query },
+    } = this.props;
     let { searchQuery } = this.props;
+    const { searchField, searchResults, eReaderUrl } = this.props;
```
EdwinGuzman: Why not merge the two destructures together?

src/app/components/SearchResults/SearchHeader.jsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
```
EdwinGuzman: It doesn't seem like you are using this file. Also, I don't think this is the best name and we try not to create layouts using `table`. So it'd be prefer if you do it with `div`s and style it rather than use tables. They should only be used for tabular data.

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -2,43 +2,42 @@ import React from 'react';
...
-  );
+export const DefinitionList = ({ work }) => {
+  /**
+   * Convert JSON object to array for parsing detail elements into
+   * a definition list for display.
+   *
+   * @param {object} work
+   * @return {string|null}
+   */
+  const workDetailsObject = workObj => Object.keys(workObj).map(key => [key, workObj[key]]);
```
EdwinGuzman: Try using `entries` instead of `map`, it'll give you what you want.

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -52,117 +51,72 @@ export const DefinitionList = ({ eReaderUrl, data }) => {
...
-                  </tr>
-                );
-              })}
-            </tbody>
-          </table>
+          <ul className="inline-list">
+            {list.map((identifier, i) => (
+              <li key={`identifiers${i.toString()}`}>
+                {identifier.id_type}
+:
```
EdwinGuzman: This is also odd.

src/app/components/WorkDetail/WorkDetail.jsx
```diff
@@ -85,14 +79,17 @@ class WorkDetail extends React.Component {
           </div>
           <div className="nypl-row">
             <div className="nypl-column-full">
-              <h2>Work Detail</h2>
```
EdwinGuzman: Will there no longer be any h2?

EdwinGuzman:
A couple of things:
1) I'm not sure where this is happening but now the footer links are blue:
Screen Shot 2019-04-08 at 6 48 42 PM

2) These _look_ like buttons so they should be `<button>`s, but they are links in the app so the are `<a>` which is okay. The problem is that you expect the whole item to be clickable and it is not. When you tab on it, only the text is focused and the whole button should be focused:
Screen Shot 2019-04-08 at 6 31 22 PM

src/app/components/Error/NotFound404.jsx
nachocs: Fixed

src/app/components/List/AuthorsList.jsx
nachocs: Fixed all of this (the formatter kept doing that)

src/app/components/List/EditionsList.jsx
nachocs: Sure, fixed.

src/app/components/SearchContainer/SearchContainer.jsx
nachocs: Fixed. Btw this comes from a eslint rule prefer-destructuring. As I saw too many errors in previous version with this, I disabled the rule. We can put it again and change all components if you think it's better.

src/app/components/WorkDetail/DefinitionList.jsx
nachocs: Fixed

nachocs:
> A couple of things:
> 
> 1. I'm not sure where this is happening but now the footer links are blue:
> 
> Screen Shot 2019-04-08 at 6 48 42 PM
> 
> 1. These _look_ like buttons so they should be `<button>`s, but they are links in the app so the are `<a>` which is okay. The problem is that you expect the whole item to be clickable and it is not. When you tab on it, only the text is focused and the whole button should be focused:
> 
> Screen Shot 2019-04-08 at 6 31 22 PM

Solved buttons and footer colors.

src/app/components/WorkDetail/DefinitionList.jsx
nachocs: The build was broken because of this, I changed it back.

src/app/components/List/AuthorsList.jsx
```diff
@@ -0,0 +1,58 @@
...
+  <span>
+    {(entity.birth_date_display || entity.death_date_display) && <span> (</span>}
+    {entity.birth_date_display && <span>{entity.birth_date_display}</span>}
+    {entity.death_date_display && <span>{` --${entity.death_date_display}`}</span>}
+    {(entity.birth_date_display || entity.death_date_display) && <span>) </span>}
+  </span>
+);
+
+const AuthorsList = ({ agents }) => (
+  <ul className="authors-list">
```
gkallenberg: We may want to set up some naming conventions unless we have some already for CSS selectors. [projectAbbrev-classname] or [nypl-projectAbbrev-className] or [nypl-className].

src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -92,27 +105,17 @@ class SearchContainer extends React.Component {
...
-                history={this.context.history}
-                {...this.boundActions}
-              />
-              <SearchResults
-                results={this.props.searchResults}
-                eReaderUrl={this.props.eReaderUrl}
-                {...this.boundActions}
-              />
+              <SearchForm searchQuery={selectedQuery} searchField={selectedField} history={history} {...this.boundActions} />
+              <SearchResults results={searchResults} eReaderUrl={eReaderUrl} {...this.boundActions} />
```
gkallenberg: Seems it's much more readable to have these as a multiline instead of inline, but if AirBnB wants it on a single line...

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -52,117 +51,62 @@ export const DefinitionList = ({ eReaderUrl, data }) => {
...
-        return Object.assign({}, item, { publisher });
-      });
-    }
+  const parseEntries = (type, entries, workObj) => {
+    const list = [...entries];
     switch (type) {
       case 'language':
         return (
           <ul>
             {list.map((entity, i) => (
```
gkallenberg: I typically prefer keys/iterators that are >= 3 characters in length, but that's just my PHP linter talking.

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -52,117 +51,62 @@ export const DefinitionList = ({ eReaderUrl, data }) => {
...
-                    </td>
-                    <td>{instance.pub_date ? instance.pub_date_display : ''} </td>
-                    <td>{instance.pub_place ? `${instance.pub_place}` : ''}</td>
-                    <td>{instance.publisher}</td>
-                  </tr>
-                );
-              })}
-            </tbody>
-          </table>
+          <ul className="inline-list">
```
gkallenberg: Prefix the className here with nypl- or similar. We need to establish a convention.

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -52,117 +51,62 @@ export const DefinitionList = ({ eReaderUrl, data }) => {
...
+            {list.map((identifier, i) => (
+              <li key={`identifiers${i.toString()}`}>{`${identifier.id_type}:${identifier.identifier}: `}</li>
+            ))}
+          </ul>
+        );
+      case 'measurements':
+        return (
+          <ul className="inline-list">
```
gkallenberg: Same here regarding className

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -52,117 +51,62 @@ export const DefinitionList = ({ eReaderUrl, data }) => {
...
+        return (
+          <span>
+            {entries}
+            {workObj.series_position && ` ${workObj.series_position}`}
+          </span>
         );
-
       default:
-        return null;
+        return _isArray(entries) ? entries.join(' ,') : entries;
```
gkallenberg: ...comma then space instead of space then comma on the `join`.

src/app/components/WorkDetail/WorkDetail.jsx
gkallenberg: This label as-is was just a placeholder until the designs were provided. The design shows the title as a header [H1] instead and subtitle as [H2], I believe. It's unclear at this point since the design document and UI/UX documentation is out of sync.

src/app/components/Application/Application.jsx
nachocs: Changed all to multiline and added eslint rule

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -172,33 +116,43 @@ export const DefinitionList = ({ eReaderUrl, data }) => {
...
+          )}
+        </tbody>
+      </table>
+    );
   };
 
-  return <dl>{getDefinitions(data)}</dl>;
+  return (
+    <div>
+      <h3 className="bold">Details</h3>
```
gkallenberg: Should have an anchor tag fragment for linking within the details view to this section of the page similar to the editions link/fragment.

src/app/components/List/EditionsList.jsx
```diff
@@ -0,0 +1,90 @@
...
+                <td className="nypl-editions-table-downloads">
+                  {instance.items ? (
+                    <EBookList
+                      ebooks={instance.items}
+                      eReaderUrl={eReaderUrl}
+                    />
+                  ) : ''}
+                </td>
+                <td>{instance.language && instance.language.length > 0 && instance.language.map(lang => lang.language).join(' ,')}</td>
+                <td>View Detail</td>
```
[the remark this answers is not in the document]
nachocs: This is in Editions, so it should be a detail about the edition, not about the work. I asked and there is no screen for view detail of edition yet

src/app/components/WorkDetail/DefinitionList.jsx
nachocs: Added anchor

PR #20, opened 2019-04-15
Added labels as provided from metadata, added language in work item

nachocs:


src/app/constants/labels.js
```diff
@@ -0,0 +1,32 @@
+export const definitionLabels = {
```
gkallenberg: We may want to change this to a more specific variable name and import it with a generic name to the DefinitionList component. We also may want to make that a somewhat more generic component itself and have an ItemDefinitionList extend that instead.

src/app/components/WorkDetail/DefinitionList.jsx
```diff
@@ -57,8 +58,8 @@ export const DefinitionList = ({ work }) => {
       case 'language':
         return (
           <ul>
-            {list.map((entity, i) => (
-              <li key={`language${i.toString()}`}>{entity.language}</li>
+            {list.map((language, i) => (
```
gkallenberg: `forEach` here instead of `map`?

nachocs: forEach wouldn't return the render values

nachocs:
> Just a couple of comments to address, plus wee seem to have lost the home page when JavaScript is turned off in the browser. One of our requirements for React apps is that they shouldn't lose their basic functionality without JavaScript. The results and item details views are working, but the home page is blank for me when trying to load in production mode. In development mode with JavaScript turned off, the styles aren't loading properly. Not sure if that's related.

that is because the front is generating index.html now, I'll remove it.

src/app/constants/labels.js
nachocs: Not sure what is your idea here, can you elaborate?

gkallenberg: Just thinking that the variable name definitionLabels for the detail's constant has a generic name and it would benefit from having a specific name like the allWorksLabels. Perhaps detailDefinitionLabels or detailLabels instead. Was just brainstorming about the component. Maybe we can chat about this today so that I have a better understanding of how components are extended in React. I'm more used to the OOP model of abstraction and extension which may not fit here.

nachocs: I changed the name of the labels

src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -93,36 +93,38 @@ class SearchContainer extends React.Component {
...
               </div>
             )}
             <div className="wrapper">
-              <SearchForm
-                searchQuery={selectedQuery}
-                searchField={selectedField}
-                history={history}
-                {...this.boundActions}
-              />
+              <div className="search-wrapper">
```
EdwinGuzman: This div is a `search-wrapper` but the other two divs above are not wrapping any search functionality or divs. A better name than `search-wrapper` would be ideal since it is wrapping the breadcrumbs and h1 and no search there.

src/app/components/Svgs/BookSvg.jsx
```diff
@@ -3,57 +3,16 @@ import React from 'react';
 
 const BookSvg = () => (
```
EdwinGuzman: This should eventually go into https://github.com/NYPL/dgx-svg-icons

nachocs: we are removing nypl libraries, so not sure about that.

src/app/components/WorkDetail/WorkDetail.jsx
```diff
@@ -61,20 +63,26 @@ class WorkDetail extends React.Component {
...
-                },
-                {
-                  href: `/work?workId=${work.uuid}`,
-                  text: 'Work Details',
-                },
-              ]}
-              pageType="details"
-              onClickHandler={handleReset}
+          <div className="search-wrapper">
+            <div className="nypl-page-header">
```
EdwinGuzman: There's no `nypl-page-header` in the search container component. Is this meant to have different html and style?

src/app/components/SearchContainer/SearchContainer.jsx
nachocs: Changed name to header-wrapper

EdwinGuzman:
This is what I see when I run the branch locally and doesn't seem right:

Screen Shot 2019-04-16 at 11 02 14 AM

src/app/components/Svgs/BookSvg.jsx
EdwinGuzman: I understand the move away from the design toolkit but these are svg icons that are agnostic to any design system. Why the move away from this?

nachocs:
> This is what I see when I run the branch locally and doesn't seem right:
> 
> Screen Shot 2019-04-16 at 11 02 14 AM

Solved. Conflict with classname in nypl

src/app/components/WorkDetail/WorkDetail.jsx
nachocs: Good point. Removed

src/app/components/Svgs/BookSvg.jsx
gkallenberg: I agree that we don't need to move icons to a different place for NYPL projects. Any of the SVG images we use should be consolidated in that repository for use by all projects. Are there standards regarding these? US.gov uses FontAwesome which I presume we could add free tier icons in the dgx-svg-icons repo, no?

PR #25, opened 2019-04-29
Pagination

nachocs:
Added pagination and per_page on search header.
Changed search query to a more generic object to be able to send it directly to back-end from store.


src/app/components/SearchResults/SearchHeader.jsx
```diff
@@ -0,0 +1,163 @@
...
+          >
+            Previous
+          </a>
+        </div>
+        <Select
+          id="page-select"
+          selectClass="sfr-select-input usa-select"
+          className="grid-col nypl-search-input"
+          options={pageList}
+          label=""
```
seanredmond: There is an empty form label on the page. Is it being generated from this empty prop? The empty label is an a11y error so we need a value here--although what would probably be better is to make the column header ("Page") the label. 

This seems to be true for all the `<select>`s

gkallenberg: Agree that form elements need to have labels and a corresponding class that hides them visually. I believe there's a default class `usa-sr-only` for hiding elements for screen readers to pick up. We might want to define that class as the default prop for form labels.

seanredmond: Stripped down we've got:

    <form>
      <div>
        <div>Items per page</div>
        <div>Sort By</div>
        <div>&nbsp;</div>
        <div>Page</div>
        <div>&nbsp;</div>
      </div>
      <div>
        <div>
          <label for="items-by-page"></label>
          <select id="items-by-page"></select>
        </div>
      </div>
    </form>

An the `<label>` is output as part of the `Select` component? Generating a label and hiding it is kind of a workaround when what I think we really mean is something like:

    <form>
      <div>
        
        <div><label for="items-by-page">Items per page</label></div>
        <div>Sort By</div>
        <div>&nbsp;</div>
        <div>Page</div>
        <div>&nbsp;</div>
      </div>
      <div>
        <div>
          
          <select id="items-by-page"></select>
        </div>
      </div>
    </form>

If we visually hide the label, then a screenreader will announce it twice: once for what's in the `<div>` and once for what's in the `<label>`

(Is this getting out-of-scope for this particular PR?)

nachocs: Made some changes so to have a label, but hidden. Anyway we must test accesibility with some tool, as with lighthouse this approach doesn't work

seanredmond:
There's a small misalignment in Firefox:
Screen Shot 2019-04-30 at 11 11 20 AM

Chrome:
Screen Shot 2019-04-30 at 11 12 29 AM

src/app/components/SearchResults/SearchHeader.jsx
```diff
@@ -0,0 +1,163 @@
...
+    const newPage = 0;
+    const newPerPage = e.target.value;
+    if (newPerPage !== searchQuery.per_page) {
+      const newQuery = Object.assign({}, searchQuery, { page: newPage, per_page: newPerPage, total: metadata.total });
+      userQuery(newQuery);
+      submit(newQuery);
+    }
+  };
+
+  return (
```
gkallenberg: The search header returns when there are no results which doesn't seem like an ideal output. Do we have designs for a no results page? Also, the pagination links show as active links when there is only 1 page. These should be disabled or removed when only one page of results is returned.

src/app/search/query.js
```diff
@@ -29,19 +29,25 @@ const parseQuery = (queryString) => {
...
+  if (_isArray(query.field)) {
+    queryField = query.field.join('|');
   }
-  const parsedQuery = parseQuery(query);
-  return { query: parsedQuery, field: queryField };
+  const parsedQuery = parseQuery(query.query);
+  return Object.assign({}, initialSearchQuery, query, { query: parsedQuery, field: queryField });
 };
 
+// export const getQueryString = query => new URLSearchParams(query).toString(); // not supported
```
gkallenberg: This should be removed.

src/app/search/query.js
```diff
@@ -29,19 +29,25 @@ const parseQuery = (queryString) => {
...
+
+export const getQueryString = query => Object.keys(query)
+  .map(key => [key, query[key]].map(encodeURIComponent).join('='))
```
gkallenberg: Encoding, yes!! :)

src/app/components/SearchForm/SearchForm.jsx
```diff
@@ -109,6 +111,7 @@ class SearchForm extends React.Component {
               value="Search"
               onClick={this.submitSearchRequest}
             />
+            <div className="grid-col-3" />
```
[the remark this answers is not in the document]
nachocs: To align it as in the design. It will be the place for "advanced search"

src/app/components/SearchResults/SearchHeader.jsx
nachocs: In the design, the search header shows when there is no results, and the pagination is still active. I will change the message on the number of items. We need to add a message in main results when there is no results according to the design, not sure if it's better to make it in other ticket

src/app/search/query.js
nachocs: URLSearchParams is more solid as it's browser implemented, but our version of node doesn't supports it yet.

src/app/search/query.js
nachocs: removing it

src/app/components/SearchForm/SearchForm.jsx
```diff
@@ -29,48 +32,46 @@ class SearchForm extends React.Component {
...
 
-    const terms = this.state.searchQuery.trim().replace(/\s+/g, ' ');
-    const encodedUserInput = encodeURIComponent(terms);
-    const path = `/search?q=${encodedUserInput}&field=${this.state.searchField}`;
+    const path = `/search?${getQueryString(this.state.searchQuery)}`;
     this.context.router.push(path);
   }
 
   render() {
+    const selectedQuery = this.state.searchQuery.showQuery || this.state.searchQuery.query || this.state.searchQuery.query;
```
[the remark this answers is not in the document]
nachocs: Removed

nachocs:
> There's a small misalignment in Firefox:
> Screen Shot 2019-04-30 at 11 11 20 AM
> 
> Chrome:
> Screen Shot 2019-04-30 at 11 12 29 AM

Fixed

src/app/components/SearchResults/SearchHeader.jsx
gkallenberg: Leave it as-is now and we'll add a new ticket for this since there may be some thought from Jack on this.

nachocs: I just added the empty results msg as in the design. Btw we should check the fonts

src/app/components/SearchResults/SearchHeader.jsx
gkallenberg: Am mainly following the guidelines here: https://v2.designsystem.digital.gov/components/form-controls/#dropdown -- We should add some label that makes sense for form inputs. Here we could add something like "Select an index" and visually hide using the Design System's default CSS class `usa-sr-only`

nachocs: Yes that's changed

nachocs: Now it works good for accesibility

nachocs: I got to make everythong 100% accesible (passing in lighthouse), changing some things in Select and TextInput

gkallenberg:
I think adding the rights information may be better suited for another PR since it doesn't fit within the pagination story.

nachocs:
> I think adding the rights information may be better suited for another PR since it doesn't fit within the pagination story.

Removed

gkallenberg:
Looks good. Just one final comment: choosing the 100 per page option only returns 20 items. 10, 20 and 50 all work as advertised. Once that's fixed it should be safe to merge. :)

nachocs:
> Looks good. Just one final comment: choosing the 100 per page option only returns 20 items. 10, 20 and 50 all work as advertised. Once that's fixed it should be safe to merge. :) Is there a way to test this on the app side?

I can't reproduce that bug. Anyway I made a commit to disable sort-by to not be confused

PR #30, opened 2019-05-08
SFR-429 no results found text need copyediting. Changed text and test

nachocs:


nachocs:
> Looks good! Out of curiosity, why a `span` and not a `div`?

Mainly because of the tests, as I needed for them something that is not a div and add a class for the test doesn't look proper. As it's just text, span is ok.

PR #32, opened 2019-05-09
SFR-391 Implement design / component.

nachocs:
After test review from Chris, this commit solve the differences found:
Added Search Footer.
Fixed some margins
First, next, previous links shouldn't be underlined
Page shown 1 of total
Format numbers
Added line down view all editions
Fix font in usa-table
Added functionality in Select to accept value and label.
SearchFooter is very similar to SearchHeader but I think it's better to keep them separated for better maintenance

### DIFF
--- a/src/app/components/SearchResults/ResultsMetadata.jsx
+++ b/src/app/components/SearchResults/ResultsMetadata.jsx
@@ -17,7 +17,7 @@ const ResultsMetadata = ({ searchQuery, metadata: { total } }) => {
     lastElement = total;
   }
   if (total > 0) {
-    message = `Viewing ${firstElement} - ${lastElement} of ${total} items`;
+    message = `Viewing ${firstElement.toLocaleString()} - ${lastElement.toLocaleString()} of ${total.toLocaleString()} items`;
   }
 
   return (
--- a/src/app/components/SearchResults/SearchResults.jsx
+++ b/src/app/components/SearchResults/SearchResults.jsx
@@ -2,8 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import ResultsMetadata from './ResultsMetadata';
 import ResultsList from './ResultsList';
-import SearchHeader from './SearchHeader';
-import SearchFooter from './SearchFooter';
+import SearchNavigation from './SearchNavigation';
 
 /**
  * Wrapper component for results lists which can contain a number of components
@@ -29,7 +28,7 @@ const SearchResults = (props) => {
           />
         </div>
         <div className="grid-col-9 sfr-results-search-header">
-          <SearchHeader
+          <SearchNavigation
             metadata={metadata}
             results={hits}
             fetchWork={props.fetchWork}
@@ -49,11 +48,12 @@ const SearchResults = (props) => {
       </div>
       <div className="grid-row sfr-results-container">
         <div className="grid-col-12 sfr-results-search-footer">
-          <SearchFooter
+          <SearchNavigation
             metadata={metadata}
             results={hits}
             fetchWork={props.fetchWork}
             {...props}
+            isFooter
           />
         </div>
       </div>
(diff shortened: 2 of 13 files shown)


src/app/components/SearchResults/ResultsMetadata.jsx
```diff
@@ -17,7 +17,7 @@ const ResultsMetadata = ({ searchQuery, metadata: { total } }) => {
     lastElement = total;
   }
   if (total > 0) {
-    message = `Viewing ${firstElement} - ${lastElement} of ${total} items`;
+    message = `Viewing ${firstElement.toLocaleString()} - ${lastElement.toLocaleString()} of ${total.toLocaleString()} items`;
```
mwbenowitz: I didn't know you could localize numbers in this way!

src/client/styles/components/EditionsList.scss
```diff
@@ -1,6 +1,6 @@
 .nypl-editions-table {
   width: 100%;
-  margin: 0 0 20px 0;
+  margin: 0;
```
mwbenowitz: The table is now running into the NYPL footer in some (all?) instances with the `mainContent` padding removed as well. Not sure where that padding should live, but it seems like there should be something there

src/client/styles/main.scss
```diff
@@ -56,7 +58,7 @@ body {
 }
 
 .main-content {
-  margin: 0 auto 60px;
+  margin: 0 auto;
```
mwbenowitz: Related to the comment above, this is causing an issue on the work detail page

src/app/components/SearchResults/ResultsMetadata.jsx
gkallenberg: Definitely a cool feature. Specific to the browser's locale? If so, do all browsers support this?

src/app/components/SearchResults/SearchResults.jsx
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import ResultsMetadata from './ResultsMetadata';
 import ResultsList from './ResultsList';
 import SearchHeader from './SearchHeader';
+import SearchFooter from './SearchFooter';
```
gkallenberg: Wondering if there couldn't be some shared components between SearchHeader and SearchFooter treatments. Are there specific things that are different now for both or would this purely be for future implementation of unique elements for this content bar? If you could share some things, that would help with minimizing repetition of code.

src/client/styles/components/EditionsList.scss
nachocs: That's true, I didn't see it. I just made a commit to solve that

nachocs:
> This looks good with just a few minor comments/questions. I left a few in the code about some `padding` issues I noticed on the work detail page.
> 
> Also a few other general thoughts:
> 
> * Did Jack ask for the new divider line between works in the search results to be heavier than the others? Can't remember what he said in standup yesterday
> * This divider line is missing when there isn't a "View all" link, is that unnecessary? unsure of what's in the designs
> * I like the footer paging, but the "Page" label sitting on its own line looks off to me. Do we need it? Obviously a minor question/one for the design team.

In the designs right now, the divider is 1 px and only shows when there is "view all". 
Yes that "Page" looks off, it's in the design like that.

src/app/components/SearchResults/SearchResults.jsx
nachocs: I began using the SearchHeader with some flag, but quickly found out that it was a very artificial  difference that could lead to confusion in the future. There are some common components that could be separated

src/app/components/SearchResults/ResultsMetadata.jsx
nachocs: Yes, it's specific to the browser locale, although we can get it fixed to 'en-US' if we want. 
Supported in all browsers

gkallenberg: No need to fix it to our way. It's always nice to see things in one's own particular idiom.

PR #33, opened 2019-05-09
SFR-408 author name not displaying

nachocs:
As we show only one author in lists and header of work item, and only the first one with the role 'author', the filter was doing it wrong.


nachocs:
> So if I understand correctly, this was `slice`ing the agent array down to one agent and _then_ filtering on the `roleFilter`? And reversing those operations is what fixed it?
> 
> If so 👍

Yes, that's it, as it was slicing it before filtering, it was showing nothing if the author was second

PR #38, opened 2019-05-14
Align search footer with head footer visually

nachocs:


nachocs:
> This looks good to me.
> 
> Another question for my own curiosity, why do we not need the `grid-col-9` in the `SearchHeader`? It stays aligned with the `SearchFooter`, but it can be trusted?
> 
> Wondering if other elements changing in the header would throw of this alignment? Perhaps that's something that can only be addressed as those changes are made.

We do have the grid-col-9, only it's in SearchResults instead

PR #39, opened 2019-05-20
Sfr 413 ebooks list

nachocs:
Using react-dropdown-aria to get a fully accessible dropdown for the ebooks, but as it requires react 16, I updated React and some associated libraries, but not react-router as it requires big changes on server side.
Also updated tests


src/app/components/List/EBookList.jsx
```diff
@@ -17,27 +19,84 @@ import PropTypes from 'prop-types';
...
+  const url = link.local && link.ebook && !link.download ? `${eReaderUrl}?url=${link.url}` : `${link.url}`;
+  const label = link.label;
+  let className;
+  if (!link.local) {
+    className = 'external';
+  }
+  return {
+    value: label,
+    url,
+    label,
```
mwbenowitz: Are `url` and `label` used by the `Dropdown` component? It seems like it only reads the other values of this object

src/app/components/List/EBookList.jsx
```diff
@@ -17,27 +19,84 @@ import PropTypes from 'prop-types';
...
-            <li key={`${ebookKey.toString()}-${linkKey.toString()}`}>
-              {generateLink(link.url, eReaderUrl, link.local, link.download, link.ebook)}
-            </li>
-          ))}
-        </ul>
-      </li>
-    ))}
+  <ul className="nypl-ebooks-list">
+    <li>{LinksSelector({ ebooks, download: true, eReaderUrl })}</li>
+    <li>{LinksSelector({ ebooks, download: false, eReaderUrl })}</li>
```
mwbenowitz: Would it be easier to filter the the ebooks here for the `download` value?

src/app/components/List/EBookList.jsx
nachocs: it's kind of duplicated, but I thought it would be better for reading what this is doing in the future. url and label are used by the onChange, and the others are used by the dropdown

src/app/components/List/EBookList.jsx
nachocs: the download value is needed for the text in the buttons

nachocs:
Updated with final design

PR #40, opened 2019-05-20
SFR-446 Use session storage. Also update library for audit

nachocs:
Synchronize Store with session storage.
Also update library for audit vulnerabilities


nachocs:
> This looks good to me. Out of curiosity how does React know when to retrieve state from session storage? Is that just automatic?

Yes the metareducer synch it automatically

PR #41, opened 2019-05-23
Sfr 415 loading

nachocs:
Added reducers and action for loading, 
Loading spinner as in design, 
interceptor for http requests, 
added a class for the store to load it from anywhere, 
removed window.onload as it's not needed, 
and changed the index.ejs template to load react after preloadedstate just in case


seanredmond:
I feel like it's out of place. My eyes are looking at the list of results, so I kind of want it centered there. (Maybe that's just centered in the window?)

Screen Shot 2019-05-23 at 11 41 38 AM

mwbenowitz:
Looks/works good to me! I'm unsure about moving it down, I think centering it over the results section makes sense, but ensuring that it remains anchored there seems much harder than fixing it in the center of the window as it is now.

seanredmond:
@mwbenowitz Do you see it centered in the window (horizontally _and_ vertically)? I see it centered horizontally but close to the top of the window

mwbenowitz:
@seanredmond You're right, it looks correct when the window is wider than it is tall, but when flipped it does appear too far to the top of the screen

nachocs:
I changed it to lower vertically

mwbenowitz:
Looks better to me in the new position. It seems to appear off screen in mobile, but I don't think we're designing/building this for mobile at this point?

Has my approval once Sean signs off on the new position and potential addition of tests?

nachocs:
Added test

nachocs:
Solved merge conflicts

PR #43, opened 2019-05-23
SFR-402 Change hierarchy in breadcrumb

nachocs:
Removed "search results" and "back to search results" link when there is no search result in the store


nachocs:
Removed search results from breadcrumbs

seanredmond:
Looks like there's no Search Results crumb when your on the search results? What we want are:

- ResearchNow
- ResearchNow / Search Results
- ResearchNow / Work Detail

Just not ResearchNow / Search Results / Work Detail

nachocs:
My apologies. Fixed now

PR #44, opened 2019-05-29
Sfr 70 show public license

nachocs:


src/app/components/List/EditionsList.jsx
```diff
@@ -21,6 +21,14 @@ const EditionsList = ({ work, eReaderUrl, max }) => {
 
   const filterValid = instances => instances.filter(instance => getIsValid(instance));
 
+  const rights = instance => instance.rights
+    && instance.rights.length > 0 //
+    && instance.rights.map(right => right.rights_statement).join(', ');
```
mwbenowitz: Generally we should only include the most recent rights statement. There are multiple rights statements so that we can track changes to the rights status of a certain work (say if it is released under a CC license but then is put into the public domain).

There is generally a `determination_date` field attached to these statements which could be used to select the most recent rights statement. However, I could also change the ElasticSearch indexing so that there should only be the most recent rights statement ever attached to these records.

Right now this should be fine since there will only be one rights statement per record, but we should consider to handle this in the future.

PR #45, opened 2019-06-03
SFR-397 Implement filters in front end.

nachocs:
Filters component, prepare query for filters
Added action for error, and displaying error in loading modal.
Make sure the booklist has unique values
Solved a warning in SearchForm
Unique Id for search header and footer
Added Filters unit test

### DIFF
--- a/src/app/components/SearchResults/Filters.jsx
+++ b/src/app/components/SearchResults/Filters.jsx
@@ -1,17 +1,18 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { isArray as _isArray } from 'underscore';
 import { initialSearchQuery, searchQueryPropTypes } from '../../stores/InitialState';
 import { getQueryString } from '../../search/query';
 
 const filtersLabels = { language: 'Language' };
 
 const Filters = ({
-  results, searchQuery, userQuery, router,
+  data, searchQuery, userQuery, router,
 }) => {
   const filtersArray = [];
 
   // add search filters
-  if (searchQuery && searchQuery.filters) {
+  if (searchQuery && searchQuery.filters && _isArray(searchQuery.filters)) {
     searchQuery.filters.forEach((filter) => {
       filtersArray.push({ field: filter.field, value: filter.value });
     });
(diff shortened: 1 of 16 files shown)


src/app/components/SearchResults/Filters.jsx
```diff
@@ -0,0 +1,139 @@
...
+  };
+
+  // update page in store and go to any page
+  const doSearchWithFilters = (filters) => {
+    const newQuery = Object.assign({}, searchQuery, { filters }, { page: 0 });
+    userQuery(newQuery);
+    submit(newQuery);
+  };
+
+  // on ckech of filter, add it or remove it from list and do the search
```
mwbenowitz: super minor typo `ckech`

src/app/components/SearchResults/Filters.jsx
```diff
@@ -0,0 +1,139 @@
...
+  const onChangeCheckbox = (e, field, value) => {
+    const matchIndex = filtersArray.findIndex(filter => filter.field === field && filter.value === value);
+    if (e.target.checked && matchIndex === -1) {
```
mwbenowitz: Is the `e.target.checked` necessary here? I only ask because it seems a bit counter-intuitive to see if the checkbox is checked to verify that we should be adding this as a new filter, at first look I thought it meant that it had already been checked and should be removed

src/app/components/SearchResults/Filters.jsx
```diff
@@ -0,0 +1,140 @@
...
+      if (!filterFound) {
+        missingFacets.push({ value: previousFilter.value, count: 0 });
+      }
+    });
+    return facets.concat(missingFacets);
+  };
+
+  // sort filters by: included in search, then count, then alphabetically
+  // then returns only the first 10
+  const prepareFilters = (facets, field) => joinFacetsAndsearch(facets, field)
```
mwbenowitz: Always appreciate a good custom `sort` method!

src/app/components/SearchResults/Filters.jsx
nachocs: Yes, as it also removes the filter when it's unchecked

PR #46, opened 2019-06-04
SFR-462 Subjects sorted alphabetically on Item Details page.

nachocs:
Removed underscore, instead using a utils library of our own to be more specific to our needs.
Sorting subjects and removed duplicates
Sorting languages by frequency of occurrence in instances
Not showing 'All Editions' if there is no instances
Added and modify tests for subjects ordering
Added class in subjects for easier testing.
Sort measurements and identifiers
Prepared in case there is no languages in instances
Updated axios version for npm audit


nachocs:
Must note there was a regression on the list of ebooks, showing only one, that it's solved with this PR

PR #47, opened 2019-06-06
Sfr 81 sort by title

nachocs:
Added tests for all links and selects of searchnavigation
sortBy Map and list of numbersByPage in separate file.
Added data-test for easier testing

### DIFF
--- a/src/app/constants/sorts.js
+++ b/src/app/constants/sorts.js
@@ -1,8 +1,9 @@
-import allowedFilters from './filters';
-
-const sorts = {
-  filter: allowedFilters,
-  order: ['asc', 'desc'],
+export const sortMap = {
+  Relevance: [],
+  'Title A-Z': [{ field: 'title', dir: 'ASC' }],
+  'Title Z-A': [{ field: 'title', dir: 'DESC' }],
 };
 
-export default sorts;
+export const numbersPerPage = [10, 20, 50, 100];
+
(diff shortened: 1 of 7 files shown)


nachocs:
I saw data-test as something commonly used, only for helping in testing. Chris suggested to add classes, but I think this approach is more solid for this purpose. Anyway here it's just a check to see what everybody think about this.

src/app/constants/sorts.js
```diff
@@ -1,8 +1,9 @@
-import allowedFilters from './filters';
-
-const sorts = {
-  filter: allowedFilters,
-  order: ['asc', 'desc'],
+export const sortMap = {
+  Relevance: [],
+  'Title A-Z': [{ field: 'title', dir: 'ASC' }],
```
seanredmond: Dash is missing in render (checked Firefox and Chrome)

Screen Shot 2019-06-07 at 10 01 03 AM

nachocs: Solved

seanredmond: Confirmed that its working for me

seanredmond:
Just to record a conversation: I agree in general with @gkallenberg about mixing testing variables and application code. In this case it isn't necessary because the element the `data-test` attribute has been added to has an `id` that @chrismulholland can use to target tests.

nachocs:
Removed data-test and updated

PR #50, opened 2019-06-13
Sfr 481 date filter

nachocs:
Added Filter for years.
Also some improvements to make search work on no-js

### DIFF
--- a/src/app/components/SearchResults/FilterYears.jsx
+++ b/src/app/components/SearchResults/FilterYears.jsx
@@ -1,23 +1,41 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { initialSearchQuery, searchQueryPropTypes } from '../../stores/InitialState';
 
 import { yearsType } from '../../constants/labels';
+
+const getYearsFilter = (searchQuery) => {
+  const yearsValues = {};
+  Object.keys(yearsType).forEach((yearType) => {
+    const yearValue = searchQuery && searchQuery.filters && searchQuery.filters.find(filter => filter.field === 'years');
+    yearsValues[yearType] = yearValue && yearValue.value[yearType] ? Number(yearValue.value[yearType]) : '';
+  });
+  return yearsValues;
+};
 
 class FilterYears extends React.Component {
   constructor(props) {
     super(props);
-    this.state = { start: '', end: '' };
+    this.state = getYearsFilter(props.searchQuery);
     this.onChangeYear = this.onChangeYear.bind(this);
   }
 
   componentWillReceiveProps(nextProps) {
-    this.setState({ start: nextProps.startYear, end: nextProps.endYear });
+    this.setState(getYearsFilter(nextProps.searchQuery));
   }
 
   onChangeYear(e, yearType) {
+    let val = e.target.value && Number(e.target.value);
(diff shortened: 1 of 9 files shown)


src/app/components/SearchResults/FilterYears.jsx
```diff
@@ -0,0 +1,89 @@
...
+    this.setState(getYearsFilter(nextProps.searchQuery));
+  }
+
+  onChangeYear(e, yearType) {
+    let val = e.target.value && Number(e.target.value);
+    if (yearType === 'start' && this.state.end && val && val > Number(this.state.end)) {
+      val = this.state.end;
+    }
+    if (yearType === 'end' && this.state.start && val && val < Number(this.state.start)) {
+      val = this.state.start;
```
mwbenowitz: These two options work well but perhaps _too_ well right now. If I enter a year into `start`, say, 2000, and then go to `end` and start typing it immediately picks up that 2 is less than 2000 and autofills 2000 from the `start` field.

I think this should wait until the number of digits in the field being entered is equal to the length of the other field before checking. Otherwise it interrupts the user's flow and forces them to go back and change the date that was entered.

src/app/components/SearchResults/Filters.jsx
```diff
@@ -83,38 +81,81 @@ const Filters = ({
...
-                  id={`facet-${field}-${facet.value}`}
-                  type="checkbox"
-                  name={facet.value}
-                  onChange={e => onChangeCheckbox(e, field, facet.value)}
-                  checked={isFilterChecked(field, facet.value)}
-                />
-                <label
-                  className="usa-checkbox__label"
-                  htmlFor={`facet-${field}-${facet.value}`}
+            <legend className="filters-box-header">{filtersLabels[field]}</legend>
```
mwbenowitz: This was giving me a syntax error that I fixed by editing this into separate rows like:
```
<legend className="filters-box-header">
  {filtersLabels[field]}
</legend>
```

src/app/components/SearchResults/FilterYears.jsx
nachocs: Not sure if the length will improve it. Maybe remove all the value changes till we have a design for errors?

mwbenowitz: That works for me, its the way it will work eventually so waiting for the design makes sense. Would that leave the min/max values in place and just remove the automatic value change?

nachocs: Removed the control over max and min. Anyway it shows an error coming from the browser, not designed, for max and min. Maybe that's enough, it looks good

nachocs:
> I'm encountering a bug where the `Filters` component is trying to render the `years` filter value as an option in the `languages` checkbox list. I think the issue lies in the `joinFacetsAndsearch` method, specifically the `find` that is happening in there. The `years` value is getting passed in and when it is not found in the `languages` list it gets added. I'm not sure the best way to fix it, nothing I tried seemed to work.
> 
> I also noticed that the year facet input fields are forcing the year to the min/max immediately after I've filled in the other value. It would be nice if it waited for me to enter a full year before checking if I've entered one too early or late.

Thank you so much for finding this bug. Solved in the new commit

PR #51, opened 2019-06-14
SFR-474: Adds title and aria-labelledby elements for decorative book svgs

helenvholmes:
After talking about it with @nachocs, I didn't add them as editable strings you can manipulate in the `<BookSvg />` element as they only refer to that particular image (seemed like overkill).


helenvholmes:
CC @mwbenowitz and @seanredmond—Pedro asked me to flag you to review this, but I can't as I'm not an official contributor.

helenvholmes:
Not certain what the protocol is on your team, but I notice this has been sitting for a day or so so I'm going to merge it. Let me know if I need to undo that/if your team has a different strategy for these for the future.

PR #52, opened 2019-06-14
SFR-466: Removes duplicate search role

helenvholmes:
Removes it from the search container to match the pattern of the search results page.


helenvholmes:
Same comment as #51.

PR #53, opened 2019-06-17
Making it work for ie11. 

nachocs:
Moved react-dropdown-aria inside the app, as it wasn't being transpiled, also added polyfills from babel, to make all work on ie11.
With this, the web works in ie11, although there are still some design differences.


nachocs:
> 👍 good fix.
> 
> I'm curious if we should be looking at other stuff in the future since `@babel/polyfill` is apparently being deprecated? The proposed replacement is confusing to me, so I'm not sure what the best approach is, but since we'll be dealing with IE11 for a while, I wonder if that might pose some challenges.


It seems they suggest to replace it to core-js, as @babel/polyfill is internally doing that, but at the same time they say we need it for webpack, so I suppose we'll wait till that is more stable.

PR #55, opened 2019-06-20
SFR-440 Windows 7 / IE 11 issues

nachocs:
The pagination boxes were not properly working for ie 11, as it have a different behavior using flex, this makes it work for ie11 and the rest of browsers

### DIFF
--- a/src/client/styles/main.scss
+++ b/src/client/styles/main.scss
@@ -149,3 +149,8 @@ select {
   -webkit-appearance: none;
   -moz-appearance: none;
 }
+
+// needed like this for ie 11
+.grid-col {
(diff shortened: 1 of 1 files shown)


nachocs:
> Looks good, would it make sense to start putting ie11 specific rules into a separate scss file? Not sure how many there might end up being.
> 
> Also the test failure seems odd, it seems to have not proper properly parsed the components, is it somehow missing the polyfill there?

Well, the css rule is not specific to IE11 in this case, it's only that IE11 doesn't understand the other form, but all browsers understand this one. In case we have some css that apply only to ie yes it would be good to have it separately.
And the build error is really odd...

PR #59, opened 2019-06-21
Adapt to new JSON response with edition_count

nachocs:
Adapt to new JSON response with edition_count property and limited intances. 
Also valid as it is now.


mwbenowitz:
@nachocs The edition_count branch has been pushed/deployed, so this should be good to deploy!

PR #63, opened 2019-06-28
SFR-495 Titles need to follow UX documentation

nachocs:
Also solved bug when going back to home that didn't remove results.
And improvement on isEmpty util that wasn't working 100%

### DIFF
--- a/src/app/components/SearchContainer/SearchContainer.jsx
+++ b/src/app/components/SearchContainer/SearchContainer.jsx
@@ -29,7 +29,6 @@ class SearchContainer extends React.Component {
 
   componentDidMount() {
     this.loadSearch();
-    this.setTitle();
   }
 
   componentDidUpdate(prevProps) {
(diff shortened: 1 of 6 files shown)


src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -29,6 +29,7 @@ class SearchContainer extends React.Component {
 
   componentDidMount() {
     this.loadSearch();
+    global.document.title = 'Search Results | ResearchNow | NYPL';
```
mwbenowitz: This is getting set when I first load the site. Is there any way that the `SearchContainer` component is mounting before the page loads and then is not rendered?

nachocs: No, but I noticed there is a mistake going back to home. Fixing it now.

src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -37,6 +38,18 @@ class SearchContainer extends React.Component {
...
     }
+    this.setTitle();
+  }
+
+  setTitle() {
+    const {
+      location: { query },
+    } = this.props;
+    if (!query || isEmpty(query)) {
+      global.document.title = 'ResearchNow';
```
mwbenowitz: I think that this should be "ResearchNow | NYPL"?

nachocs:
I refactored all this, using a common const and applying the title in the Application instead of components, please review again

src/app/components/SearchContainer/SearchContainer.jsx
nachocs: That's not approved yet, I think

PR #64, opened 2019-07-01
SFR-502 View all shortcut only works visually.

nachocs:
Added tabindex=-1

### DIFF
--- a/src/app/components/List/EditionsList.jsx
+++ b/src/app/components/List/EditionsList.jsx
@@ -78,7 +78,10 @@ const EditionsList = ({ work, eReaderUrl, max }) => {
       </table>
       {!!max && editionCount > max && (
         <div className="nypl-editions-view-all">
-          <Link to={{ pathname: '/work', query: { workId: `${work.uuid}` }, hash: '#all-editions' }}>
+          <Link
+            to={{ pathname: '/work', query: { workId: `${work.uuid}` }, hash: '#all-editions' }}
+            tabIndex="-1"
+          >
             View All&nbsp;
             {editionCount}
             &nbsp;Editions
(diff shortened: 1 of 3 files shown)


nachocs:
I agree, I think it should be able to go by tabs, but that's what Accesibility wanted...

mwbenowitz:
> I agree, I think it should be able to go by tabs, but that's what Accesibility wanted...

In further testing it seems like the focus is still staying on "View All....", I tried experimenting with this but I can't see how to force the focus to move without resorting to React, `tabIndex` seems to have no effect, unless I'm testing this incorrectly/this is a browser-specific issue

seanredmond:
There are a few problems here. First the tabindex needs to be on the _target_ element, not the "View all editions" link. So the attribute needs to move from the link to the "All editions" header

Second, we don't need the anchor with no href just to be the target of `#all-editions` (and it seems to me that a link with no href complicates keyboard focus). Just put the id on the h3, along with the tabindex attribute, like so:

    <h3 id="all-editions" tabIndex="-1" className="all-editions-tag bold">
        All Editions
    </h3>

Finally, I don't think these changes quite achieve the goal. It seems counter-intuitive to me but, at least with VoiceOver, following an internal link still doesn't move the focus, even with the tabindex. If you tab to the "View All" link and activate it, the screen jumps to the expected heading, but if you tab again you jump back up the the next link _after the "View All" link_--not the next link after the "All Edition" header.

I think we need a handler on the "View all editions" link to to explicitly `focus()` the `#all-editions` element. This seems to make things work as expected.

nachocs:
Changed as we talked. Also fixed a double load when linking with the anchor

nachocs:
With the previous solution the tag didn't follow index. Making if go to focus and adding tabindex works the trick... though it's not recommended for accesibility... let me know your thoughts

nachocs:
Now it works as expected :)

nachocs:
I removed the hashtag scroller from the App to the component, to it will scroll and focus when the data is loaded, take a look.

PR #71, opened 2019-07-24
SFR-542 Show advanced search

nachocs:


src/app/components/SearchResults/AdvancedSearchResults.jsx
```diff
@@ -0,0 +1,60 @@
...
+import { initialSearchQuery, searchQueryPropTypes } from '../../stores/InitialState';
+import { getQueryString } from '../../search/query';
+
+const AdvancedSearchResults = ({ searchQuery, userQuery, router }) => {
+  // redirect to url with query params
+  const submit = (query) => {
+    const path = `/search?${getQueryString(query)}`;
+    router.push(path);
+  };
+  // update page in store and go to any page
```
mwbenowitz: I really like having this functionality, but I don't think it was in the scope/ticket for this work?

The only reason that I raise this is that it allows the user to clear _all_ queries, resulting in the return of every work in the database. There isn't anything wrong with this, but it will put them in a situation that I hadn't anticipated and things such as paging will likely break if a user tries to take actions from this state.

I don't think this should be removed, since it is functionality we will almost certainly want in the future, but perhaps, the buttons could be disabled for now?

nachocs: I thought this is what is expected from this kind of tags, otherwise there is no way to remove advanced search terms

nachocs: Anyway now you can get to full results, just search in advanced search for nothing with a language, then remove the language.

mwbenowitz: With the discussion we just had I think it makes sense to make them static. I agree that its the generally expected behavior, but it just intersects with too many other decisions about how we return results to be activated right now.

src/app/components/AdvancedSearch/AdvancedSearch.jsx
```diff
@@ -207,15 +206,32 @@ class AdvancedSearch extends React.Component {
...
+      return;
+    }
     if (fullQuery) {
       this.boundActions.userQuery(fullQuery);
       const path = `/search?${getQueryString(fullQuery)}`;
       this.context.router.push(path);
     }
   }
 
+  validate() {
```
mwbenowitz: Does this method get called anywhere? Seems the functionality is built in at 209-212

PR #74, opened 2019-07-30
NYPL Header

nachocs:
Added NYPL Header as a external library


nachocs:
> This looks really good, really nice to be able to include it as an external dependency.
> 
> The only issue I see is in the "Get email updates" dropdown, the dropdown doesn't seem to be sizing properly due to a `box-sizing` issue. Setting this to `content-box` fixes it, and it seems to be due to some kind of inherit issue from the footer? I'm not actually sure about the source.
> 
> Its a minor issue if we want to ignore it for now, just wanted to raise it as something I noticed.

Easy solved. Commited

PR #77, opened 2019-08-05
Refactor of Filter Years to make it work in Filters and Advanced Filter

nachocs:


src/app/components/SearchResults/FilterYears.jsx
```diff
@@ -16,62 +17,84 @@ const getYearsFilter = (searchQuery) => {
...
   render() {
     return (
       <div className="grid-container padding-0">
-        <div className="grid-row">
+        <div className={this.props.className}>
           {Object.keys(yearsType).map(yearType => (
-            <label
+            <TextInput
+              className={this.props.inputClassName}
+              ariaLabel="Search for End Date"
```
mwbenowitz: I think the aria-label should be tied to the `yearType`? It seems to be outputting "End" for both the start and end years.

nachocs: Totally right. Just changed

PR #80, opened 2019-08-23
SFR-578: Item with no links causes error

hokei:
Bugfixes:
- Filter out items with no links
- Don't re-do search when results are available
- Don't call redux ERROR action when there is no error

### DIFF
--- a/src/app/components/SearchContainer/SearchContainer.jsx
+++ b/src/app/components/SearchContainer/SearchContainer.jsx
@@ -44,7 +44,8 @@ class SearchContainer extends React.Component {
     const {
       location: { query },
       dispatch,
-    } = this.props;
+      searchQuery,
+    } = this.props;    
 
     if (!query || isEmpty(query)) {
       this.boundActions.resetSearch();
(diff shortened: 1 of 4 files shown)


src/app/components/SearchContainer/SearchContainer.jsx
```diff
@@ -59,7 +60,7 @@ class SearchContainer extends React.Component {
       if (query && query.queries) {
         newQuery = Object.assign({}, newQuery, { queries: JSON.parse(query.queries) });
       }
-      if (query) {
+      if (searchQuery && !deepEqual(newQuery, searchQuery)) {
```
mwbenowitz: On closer inspection I don't actually think this is where the unnecessary `searchPost` action is coming from! I'm trying to figure it out, but I think one of the reducers is picking it up and executing it? Can demonstrate what I've found so far.

mwbenowitz: This seems to have been the result of a flubbed merge by me. Can't reproduce after resetting to `HEAD`. Will close this comment.

PR #82, opened 2019-09-03
SFR-581 Bugfixes

hokei:
- Filter Refactor to isolate functions
- Error validation for sidebar filter years


test/unit/Filters.test.js
```diff
@@ -103,4 +105,48 @@ describe('Filters', () => {
       ).to.equal(true);
     });
   });
+
+  describe('Filter Interactions', () => {
```
hokei: Still working on the "show error when dates are wrong" test, but I thought I'd push it anyway, this situation is proving a bit tricky to test in an automated way due to the way enzyme's simulate works.

PR #83, opened 2019-09-10
SFR-582 Clean Up React Warnings

hokei:
Changed link dropdown label


hokei:
@mwbenowitz  I think it doesn't hurt to leave it for debugging - especially since the way our Dropdowns work, it uses the labels as keys.  There's a lot of good info there which will still come in handy for debugging, especially if an error gets thrown from there.

PR #87, opened 2019-09-16
SFR-381 Added Work count controlled by feature flag

hokei:
Screen Shot 2019-09-16 at 4 29 35 PM

### DIFF
--- a/src/app/components/SearchForm/TotalWorks.jsx
+++ b/src/app/components/SearchForm/TotalWorks.jsx
@@ -1,0 +1,51 @@
+import React from 'react';
+import { bindActionCreators } from 'redux';
+import { connect } from 'react-redux';
+
+import PropTypes from 'prop-types';
+import * as searchActions from '../../actions/SearchActions';
+
+
+class TotalWorks extends React.Component {
+  constructor(props) {
+    super(props);
+    const { dispatch } = props;
+    this.boundActions = bindActionCreators(searchActions, dispatch);
+  }
+
+  componentDidMount() {
+    this.getBookTotals();
+  }
+
+  getBookTotals() {
+    this.props.dispatch(searchActions.fetchTotalWorks());
+  }
+
+  render() {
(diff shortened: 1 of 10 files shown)


src/app/components/SearchForm/TotalWorks.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+
+class TotalWorks extends React.Component {
+  constructor(props) {
+    super(props);
+    const { dispatch } = props;
+    this.boundActions = bindActionCreators(searchActions, dispatch);
+  }
+
+  componentDidMount() {
```
mwbenowitz: Should this be `componentWillMount`? I was getting errors that `this.props.totalWorks` was `undefined` in the `render` method

src/app/components/SearchForm/TotalWorks.jsx
```diff
@@ -0,0 +1,51 @@
...
+  totalWorks: PropTypes.objectOf(PropTypes.any),
+  dispatch: PropTypes.func,
+};
+
+TotalWorks.defaultProps = {
+  totalWorks: { data: { counts: { works: 0 } } },
+  dispatch: () => { },
+};
+
+export default connect(mapStateToProps, null)(TotalWorks);
```
mwbenowitz: The component is not appearing for me when I load the search form. I get a warning, but I don't think its related.

I could be very wrong about this, but if you don't export the `TotalWorks` object how does the app load it?

src/app/components/SearchForm/TotalWorks.jsx
mwbenowitz: Apparently a cache/storage issue, marking as resolved

src/app/components/SearchForm/TotalWorks.jsx
mwbenowitz: Same as above

PR #89, opened 2019-09-19
SFR-583 Moved Readers into iFrame

hokei:
Also did an npm update to get rid of some deprecated dependencies.  It doesn't look like it's changed anything, but let's keep an eye out for that!


src/app/util/Util.js
```diff
@@ -57,10 +57,28 @@ export const checkFeatureFlagActivated = (featureFlagList, componentStateObject)
   });
 };
 
+// Given a link, return the link with 'http' and strip https if present
+export const formatUrl = (link) => {
```
mwbenowitz: This isn't necessary at all but I think this could be simplified using a regex. But I tend to abuse regexes so it is entirely up to you
```
if (link.match(/^http[s]?:\/\//)) { return `http://${link}` }
return link
```

package.json
```diff
@@ -94,12 +94,12 @@
...
+    "node-sass": "^4.12.0",
+    "optimize-css-assets-webpack-plugin": "^5.0.3",
     "prop-types": "15.6.2",
-    "react": "^16.8.6",
+    "react": "^16.9.0",
     "react-a11y": "1.1.0",
-    "react-dom": "^16.8.6",
+    "react-dom": "^16.9.0",
     "react-redux": "5.1.1",
     "react-router": "3.0.5",
```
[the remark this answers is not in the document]
hokei: Yeah, it's still fine.  It's that "componentWillMount is deprecated" thing - it will still work until react 17.  I think I'll just take a task to re-write all the deprecated methods sometime soon!

src/app/util/Util.js
hokei: Regex totally appeals to the "brevity and elegance" part of my brain, but I struggle with writing and reading it, so I'm gonna leave it

PR #91, opened 2019-10-03
Add to changelog and bump version

hokei:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "rnw-front-end-development",
-  "version": "0.0.1",
+  "version": "0.3.1",
```
hokei: This wasn't done in the past ... is this right?

mwbenowitz: It is, I always forget to bump the version there

PR #92, opened 2019-10-10
SFR-602 converted read-online links to webpub viewer

hokei:


src/app/components/List/EBookList.jsx
```diff
@@ -62,25 +86,37 @@ const arrowRenderer = (open) => {
...
-const linksArray = ({ ebooks, download }) => unique(
-  flattenDeep(ebooks.filter(item => item.links).map(item => item.links)) //
-    .filter(link => link.download === download), //
-  'url',
-);
-// const linksArray = ({ ebooks, download }) => flattenDeep(ebooks.map(item => item.links)).filter(link => link.download === download);
+// gets all links from all ebooks
+// First searches for eReader links
+// If it can't find then, it classifies them on download
+// TODO: This method does not work on items that have both eReader and non-eReader links
```
hokei: Just want to make sure this is an OK assumption to make, especially for now.

mwbenowitz: I think its an okay assumption to make for the time being, but I think there is a way to implement this so that we can combine the two. Really what we want to filter out are the `content.opf` links, which aren't used right now, right?

The simplest logic to check for that seems to be `link.ereader === true || link.local === false`, which should filter out non-ereader local files, if I'm thinking about this correctly. Then you could just return that array without having the extra `flattenDeep` call and get both types of links together.

Not 100% sure about that and we can obviously put this off for another branch/ticket if we feel that it merits closer attention or a change to the flags that are getting set.

mwbenowitz: Wait that logic would actually be `link.ereader === true || (link.local === false && link.download === download)`, which is probably a sign that this merits more thought

src/app/components/List/EBookList.jsx
```diff
@@ -19,9 +32,11 @@ import { flattenDeep, unique } from '../../util/Util';
  * @param {boolean} ebook
  * @return {string}
  */
-const generateLink = (url, eReaderUrl, local, download, ebook) => {
-  const link = local && ebook && !download ? `${eReaderUrl}?url=${url}` : `${url}`;
+const generateLink = (url, eReaderUrl, local, download, ereader, ebook) => {
+  const encodedUrl = generateStreamedReaderUrl(url, eReaderUrl);
+  const link = local && ebook && ereader && !download ? `${encodedUrl}` : formatUrl(url);
   if (ebook && !download) {
+    console.log('got here', link);
```
mwbenowitz: Just extra console.log statement

src/app/components/List/EBookList.jsx
```diff
@@ -2,7 +2,20 @@ import React from 'react';
...
+/**
+ * Create a URL for that conforms to the webpub-reader streamed format
+ * @param {string} url
+ * @param {string} eReaderUrl
+ */
+const generateStreamedReaderUrl = (url, eReaderUrl) => {
+  const base64BookUrl = Buffer.from(formatUrl(url)).toString('base64');
+  const encodedBookUrl = encodeURIComponent(`${base64BookUrl}`);
+  const encodedReaderUrl = encodeURI(eReaderUrl);
+  const combined = `${eReaderUrl}/readerNYPL/?url=${encodedReaderUrl}/pub/${encodedBookUrl}/manifest.json`;
```
mwbenowitz: How does this work with the `/manifest.json` at the end? I don't think its in the .epub files, right? It does work, I'm just curious what its doing.

hokei: I think this is a result of the streamer, because what it's doing right now is taking the unexploded epubs and running it through the readium streamer so that it can be displayed to us.

mwbenowitz: Okay, I think I understand? So its reading the generated `manifest.json` file, right? Or does it just expect that to be there and ignores it in what is actually being parsed?

hokei: When the epub is streamed through the reader, it's made available through a link that's essentially [host]/[base_64_encoded_direct_link]/manifest.json 
So I don't think it'll be found if the manifest.json is stripped (like, url-wise, cuz it's not pointing to the right file anymore)

All we're doing is passing that url into RNow to show in an iframe.

hokei:
@mwbenowitz I think the only place the reader isn't being used, but `generateStreamedReaderUrl` is being called, is in the option generation, and I don't think it hurts to have it there.  Buffer is just encoding the url as if it's a string, so if we pass it a weird one, it'll just encode that too, but then the e-reader won't show the book - I think, that's unavoidable and also ok.

mwbenowitz:
@hokei ok great, was unclear on what the errors there could be but that sounds good!

PR #94, opened 2019-10-11
SFR-595 Add ga tracker for pages

mwbenowitz:
As a single-page application ResearchNow requires a little additional configuration to track page views, and through that, search events. This can be done by hooking into the history object and listening to changes to the history, which can fire the ga tracker. With these events in ga the site search features should work automatically.

This can't be tested locally but I `console.log`'d the output of `location.pathname` on page load and it does fire properly on page changes/loads with a path that GA should understand

### DIFF
--- a/src/client/App.jsx
+++ b/src/client/App.jsx
@@ -2,6 +2,7 @@ import '@babel/polyfill';
 import React from 'react';
 import a11y from 'react-a11y';
 import ReactDOM from 'react-dom';
+import { gaUtils } from 'dgx-react-ga'
 import FeatureFlags from 'dgx-feature-flags';
 import { Router, browserHistory } from 'react-router';
 import { Provider } from 'react-redux';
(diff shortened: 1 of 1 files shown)


hokei:
This looks good! Interested in whether the analytics may end up with some interesting results (the path sometimes shows weird things when using the back button, the history object sometimes persists state when it shouldn't)  It might affect this, it might not - I'd click around on the app after this is merged and see what that analytics records, this might be a good push to fixing the history object (and be a good test on whether it is indeed fixed!)

mwbenowitz:
@hokei ok, thanks for the heads up! There are apparently multiple ways to add ga tracking to single page apps, this just seemed like the cleanest way that should only trigger once per page "load". If it reports weird results I agree we should try to look at what the root cause is and see if we should change our analytics approach or how history is tracked.

PR #96, opened 2019-10-23
SFR-668 removed 'query' and 'keyword' parameters in favor of 'queries'

hokei:
Updated tests as well
Not really sold on this approach, (the hardcoding of the first entry in `queries`) but it's working ok.  Happy to take another look at it now, or defer to when we have new designs!

### DIFF
--- a/src/app/stores/InitialState.js
+++ b/src/app/stores/InitialState.js
@@ -1,8 +1,6 @@
 import PropTypes from 'prop-types';
 
 export const initialSearchQuery = {
-  query: '',
-  field: 'keyword',
   showQuery: '',
   showField: '',
   per_page: 10,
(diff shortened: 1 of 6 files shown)


src/app/stores/InitialState.js
```diff
@@ -1,21 +1,17 @@
 import PropTypes from 'prop-types';
 
 export const initialSearchQuery = {
```
mwbenowitz: I think the simplification of this part alone justifies making this change, so much clearer!

hokei:
Yeah, it was being populated based on a write to state, but I kinda like it more like this than if it didn't populate at all (especially since there isn't a way to remove the "chicklets" so far)

PR #97, opened 2019-11-05
SFR-702 Removed Kievit and replaced with system fonts

hokei:
Ripped out Kievit.  
I did a cursory search for it while running in development, and didn't find it anymore.


hokei:
Question:  Should I be updating the changelogs/versions with every commit? This should probably be deployed after merging, but I also haven't been updating changelogs for previous things.

src/client/styles/components/EBookList.scss
```diff
@@ -6,7 +6,8 @@ ul.nypl-ebooks-list {
   color: #0071bb;
   padding: 0;
   font-size: 1rem;
-  font-family: 'Kievit-Bold';
+  font-family: system-ui;
```
helenvholmes: Because you have this declared in `main.scss` as `$fontFamily` applied to the html element globally, you shouldn't need these additional `font-family` declarations. I'd pull these out, since that was the point of that abstraction anyway—that you can make global font-family changes relatively quickly.

src/client/styles/components/EBookList.scss
```diff
@@ -43,7 +44,8 @@ ul.nypl-ebooks-list {
     background-size: 1rem;
     background-position: right 0.5rem center;
     font-size: 1rem;
-    font-family: 'Kievit-Bold';
+    font-family: system-ui;
```
helenvholmes: Please delete

src/client/styles/components/EBookList.scss
```diff
@@ -62,15 +64,17 @@ ul.nypl-ebooks-list {
   text-align: center;
   text-decoration: none;
   font-size: 1rem;
-  font-family: 'Kievit-Bold';
+  font-family: system-ui;
```
helenvholmes: Please delete

src/client/styles/components/EBookList.scss
```diff
@@ -62,15 +64,17 @@ ul.nypl-ebooks-list {
...
+  font-weight: 700;
   ul {
     width: 11.5rem;
     white-space: normal;
     border: 2px solid black;
     border-radius: 8px;
     padding: 0;
     button {
-      font-family: 'Kievit-Book';
+      font-family: system-ui;
```
helenvholmes: Please delete

src/client/styles/components/Filters.scss
```diff
@@ -1,7 +1,8 @@
 .filters {
   padding: 1rem 0 1rem 2rem;
   .filters-header {
-    font-family: 'Kievit-Bold';
+    font-family: system-ui;
```
helenvholmes: Please delete

hokei:
@helenvholmes It looks big, but it isn't Kievit.  Will merge, I think it's ok to fix this with the design system updates. 
Screen Shot 2019-11-06 at 11 25 55 AM

PR #98, opened 2019-11-18
SFR-616 incorporate breadcrumb

hokei:
Adds Design System Breadcrumbs into ResearchNow


src/app/components/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -10,49 +11,31 @@ import { Link } from 'react-router';
...
-    <div className="grid-row">
-      <nav
-        aria-label="Breadcrumbs"
-        className="grid-col-10 sfr-center nypl-breadcrumbs"
-      >
-        <span className="nypl-screenreader-only">You are here:</span>
-        <ol>{crumbs}</ol>
-      </nav>
-    </div>
+    <DS.Breadcrumb breadcrumbs={crumbs} />
```
hokei: I'm a bit torn on this naming - On one hand, I think it's better to allow products to wrap DS components using their own functionality, shown in this Breadcrumbs class.  But it does mean that we want to namespace DS components.

mwbenowitz: I absolutely think this is the way to go, it makes it clear where this component is coming from and it is important to make a distinction between the design components and the local components implementing them.

PR #99, opened 2019-11-21
Removed current page from breadcrumb

hokei:


src/app/components/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -23,10 +23,12 @@ const Breadcrumbs = ({ links, pageType, onClickHandler }) => {
     const crumbs = [homeLink];
 
     if (links && links.length && pageType !== 'home') {
-      links.forEach((link) => {
-        crumbs.push(
-          <Link to={link.href}>{link.text}</Link>,
-        );
+      links.forEach((link, i) => {
+        if (i < links.length - 1) {
```
seanredmond: Would it not be snazzier to use something like an array slice here? What happens if no Links are added?

hokei: Yup, that would be cleaner!  If no links are added it displays `homeLink` from L23, though it's a good point in that DS.Breadcrumbs should probably throw an informative error when no links are passed in - that would be helpful for developers!

PR #100, opened 2019-11-22
SFR-616 Clean-up

hokei:
Did some more clean up.

### DIFF
--- a/src/app/components/Breadcrumbs/Breadcrumbs.jsx
+++ b/src/app/components/Breadcrumbs/Breadcrumbs.jsx
@@ -10,34 +10,35 @@ import * as DS from '@nypl/design-system-react-components';
  * @returns {array}
  */
 const Breadcrumbs = ({ links, pageType, onClickHandler }) => {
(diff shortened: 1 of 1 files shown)


src/app/components/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -10,34 +10,35 @@ import * as DS from '@nypl/design-system-react-components';
...
     if (links && links.length && pageType !== 'home') {
-      links.forEach((link, i) => {
-        if (i < links.length - 1) {
-          crumbs.push(
-            <Link to={link.href}>{link.text}</Link>,
-          );
-        }
-      });
+      // If not on home, add all the other links, except for the current page
+      crumbs.concat(links.slice(0, links.length - 1)
```
seanredmond: You can just write `links.slice(0, -1)`

hokei: did not know that!  Nifty!

PR #101, opened 2019-11-27
SFR-616 Removed hover text-decoration

hokei:
In RN the breadcrumb wasn't showing an underline on hover, because this was over-riding it.


hokei:
As discussed in stand up - This seems to be overridden everywhere else, and this change hasn't affected anything other than the breadcrumbs.

PR #103, opened 2019-12-16
SFR-770 webpub back button

hokei:


appConfig.js
```diff
@@ -6,7 +6,7 @@ export default {
   webpackDevServerPort: 3000,
   baseUrl: '',
   ereader: {
-    development: 'http://researchnow-webpub-dev.us-east-1.elasticbeanstalk.com:4444',
+    development: 'http://localhost:4444',
```
hokei: Will revert this before merge, though, thinking about setting up a flag or something to ease development.  Might do it in the next viewer ticket?

src/app/components/Viewer/EBookViewer.jsx
```diff
@@ -14,6 +14,21 @@ class EBookViewer extends React.Component {
 
   componentDidMount() {
     this.parseQueryToState(this.props.location.query.url);
+    this.handleIframeTask = this.handleIframeTask.bind(this);
+    window.addEventListener('message', this.handleIframeTask, false);
+  }
+
+  handleIframeTask(e) {
+    if (e.origin !== 'http://localhost:4444') {
```
hokei: change to aws before merge

src/app/components/Viewer/EBookViewer.jsx
```diff
@@ -14,6 +14,21 @@ class EBookViewer extends React.Component {
...
+      return;
+    }
+    if (e.data === 'backButtonClicked') {
+      // TODO: if (previous location within ResearchNow app) {
```
hokei: This check is not working, and I'm gonna table it for now.  I'm not sure if it's worth it to figure out in react-router-v3 when we're probably gonna upgrade to react-router-4+ in a couple months.

mwbenowitz: What was this attempting to check? Going back to the last link in the app would be a really nice feature.

hokei: So React-Router's `goBack()` will take you to the last location, but it requires a well set-up history object to be able to check where that is.  Our history object and our router objects are mixed up, I think we're using the `router` object as `history` and `history` is always `undefined`.  I'd love to fix this, but I think part of that fix is going to be upgrading `react-router` because trying to search the internet on info on `react-router@3` is a real pain.

hokei:
@mwbenowitz I'm gonna try some other ideas -- I agree that having that button mirror the browser's `back` is undesirable.

appConfig.js
mwbenowitz: A flag to be able to set this for local development (or to provide zero-downtime transitions between readers) seems like a really good idea

hokei:
@mwbenowitz I think this will do what we want better.

hokei:
@mwbenowitz OK hope this works.  
I've changed the "read online" and "download" items into `buttons` instead of `Link`s so that everything can use the same code path.  New designs will have these be buttons, too, so uh it's worse looking for now but it's better this way.

mwbenowitz:
@hokei The fix works and I agree that the visual design doesn't matter as that aspect will be refreshed shortly.

The only thing I'd like to raise is that I believe that this was a `Link` because there were accessibility concerns with using a button for page navigation. However I think this will be also addressed by the new designs, just want to bring it up to be sure we don't go down a path that we will have to walk back later.

I would merge this though!

hokei:
Yeah, I think it'll be ok if we put the `Link` back in as a part of the results page template work! 
~merge merge merge~

PR #104, opened 2019-12-17
SFR-603 Landing Page

hokei:
New Landing Page!  
It should already be pointing to the branch, but, just in case -- this is intended to work with 
https://github.com/NYPL/nypl-design-system/tree/SFR-603_ResearchNow_Landing! 

The main difference is that we no longer should load search results/advanced search on the `SearchContainer.jsx` page - the reason is that the search results pages have a different looking searchbar, so we should use those components.  However, I've left the SearchResults component in to verify that searching still works.


hokei:
@helenvholmes 
Screen Shot 2019-12-17 at 4 23 39 PM

Some of the CSS is off (I think, because conflicting styles), especially on a very wide screen.  
Screen Shot 2019-12-17 at 4 25 09 PM

At least one of these is an error carried over from design system:  (SearchPromo doesn't expand properly when the page expands) 
Screen Shot 2019-12-17 at 4 26 20 PM

You can see it yourself by pulling this repo and running `npm i` and `npm run dev-start`!

hokei:
To Do:  

- [ ] Tests
- [ ] Search query doesn't seem to visually reset on page refresh
- [ ] Error on clicking the last subject

src/app/components/SearchForm/SearchForm.jsx
```diff
@@ -121,8 +122,37 @@ class SearchForm extends React.Component {
     const selectedField = this.state.searchQuery.showField || this.state.searchQuery.queries[0].field;
 
     return (
-      <div className="grid-row">
-        <form
+      <DS.SearchPromo
+        headingText="Search the World's Research Collections"
```
hokei: This whole series of props looks very hard to read to me, and I'm wondering if there's a better way to order/sort/put them in `{...inputprops}, {...searchBarProps}, {...paragraphProps}` or something.  

@EdwinGuzman @helenvholmes does this bother you as well? Thoughts?

EdwinGuzman: I prefer one prop per line. I'd also make the `advancedSearchMessage` value into a variable. If it's easier for _you_ to combine into a variable and explode it, then go for it.

(Some package conventions like prettier opt for one prop per line, too).

hokei:
@helenvholmes Aaaaaahhh it's very close!  
major things I notice:  

- [ ] In the Design System and in Research Now, the image-right gets smaller as you make the window smaller (until it gets to mobile size).  I remember we had a conversation about this earlier and you had mentioned needing some sort of container to put header-image-right in.  Is that what's missing here?  
- [ ] The colour for the breadcrumb and for the ResearchNow header is different (http://themetronome.co/templates/t-51-landing-page/?fresh=true) but in the current view, it looks like it's the same colour.  Where do these colours live / can we update them in the DS for now? 
- [ ] There's a weird space between the breadcrumb and the page title.  This is due to `normalize.css` overriding the `_headings.scss` mixin ... `normalize` isn't imported anywhere inside ResearchNow, but in Design System, it seems to work ok via.  I'm pretty stumped on this one, do you think it has anything to do with the way mixins might get resolved?

helenvholmes:
> In the Design System and in Research Now, the image-right gets smaller as you make the window smaller (until it gets to mobile size). I remember we had a conversation about this earlier and you had mentioned needing some sort of container to put header-image-right in. Is that what's missing here?

This is something for the life of me I can't make any better than it is now. I've throw in some extra media queries in a branch on the DS so it isn't so jarring, but both 1) maintaining aspect ratio on the image as the browser resizes while also having the image stretch to full height is really tricky and 2) getting the green part to expand past the image is also rough when it's laid out with flexbox. Bleh.

> The colour for the breadcrumb and for the ResearchNow header is different (http://themetronome.co/templates/t-51-landing-page/?fresh=true) but in the current view, it looks like it's the same colour. Where do these colours live / can we update them in the DS for now?

I've updated `_color_vars.scss` (where the variables `$nypl-turquoise-regular` and `$nypl-turquoise-dark` live) to be the new colors. I had originally put in Jack's original colors and they've changed significantly for ResearchNow (can we even call that turquoise? Looks kinda seagreen to me, idk). I've added it as something to discuss in [Housekeeping](https://docs.google.com/document/d/16C5ZCAkTNt_n75FzDqvueaVjlVgDh9ye6C7O62z_yLo/edit?ts=5de6c34c#) tomorrow. 

> There's a weird space between the breadcrumb and the page title. This is due to normalize.css overriding the _headings.scss mixin ... normalize isn't imported anywhere inside ResearchNow, but in Design System, it seems to work ok via storybook. I'm pretty stumped on this one, do you think it has anything to do with the way mixins might get resolved?

It seems like `normalize` is getting injected into the header in a `<style>` tag in ResearchNow. Do you know where that's getting injected? The Design System isn't doing that... is it?

[Here is the now badly named PR on DS that fixes points 1&2](https://github.com/NYPL/nypl-design-system/pull/66).

hokei:
> It seems like normalize is getting injected into the header in a <style> tag in ResearchNow. Do you know where that's getting injected? The Design System isn't doing that... is it?

Ugh.  Thanks for finding it, I'll go looking for it again

PR #105, opened 2019-12-27
SFR-761 fix subject query handling

mwbenowitz:
This fixes several issues around the handling of queries. The initial ticket was to investigate `500` errors stemming from making subject queries from the work detail pages. That seems to have been fixed by a different change, however investigating it revealed that non-keyword searches were not being executed properly and that the subject searches from the detail pages were not properly displaying the query "pills".

Both issues were due to their respective components using the old-style simple search query structure. Updating both of these to the new advanced search structure has fixed both issues.

The one point that could use further review is coming from the detail page I had to encode the query array as a string to get the app to parse it correctly, otherwise it rendered it as a string. I'm not sure if there was an issue with how I was passing the data, or if there is something with how that component understands the parameters passed to it, but it could use a second set of eyes.

### DIFF
--- a/src/app/components/SearchForm/SearchForm.jsx
+++ b/src/app/components/SearchForm/SearchForm.jsx
@@ -81,7 +81,7 @@ class SearchForm extends React.Component {
     this.setState((prevState) => {
       const advancedQuery = {
         query: querySelected,
-        field: prevState.searchQuery.showField ? prevState.searchQuery.showField : prevState.searchQuery.field || 'keyword',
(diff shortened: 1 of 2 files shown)


hokei:
I think there's something wrong in general with how we're serializing/deserializing url parameters (eg: if you go to `homepage/blah` it'll also not parse).  I think this is worth filing a separate ticket for, and I'll make a note of this particular workaround in that ticket.  Otherwise, this looks good to me!

PR #108, opened 2020-01-16
SFR-442 add feedback form

mwbenowitz:
This adds a feedback button for user bug reports/responses to the ResearchNow app. The base for this form was borrowed from Shared Collections Catalog and modified with new styles, a new question and a new integration. It has also been updated to work with our version of React, which is somewhat newer than what is in use on SCC.

This sends the responses to an Airtable spreadsheet, which is easier to work with than a Google Sheet Integration.

This has only a very basic test and I'm not sure if the best thing to do is to add more tests now or have some basic tests in place and build out the test suite in the future.

### DIFF
--- a/src/app/components/Feedback/Feedback.jsx
+++ b/src/app/components/Feedback/Feedback.jsx
@@ -13,7 +13,7 @@ class Feedback extends React.Component {
       showForm: false,
       feedback: '',
       success: null,
-      commentStatus: '',
+      commentStatus: false,
     };
 
     this.feedbackField = React.createRef();
@@ -32,7 +32,7 @@ class Feedback extends React.Component {
     e.preventDefault();
     if (!this.state.feedback) {
       this.feedbackField.focus();
-      this.setState({ commentStatus: 'Comment Required' });
+      this.setState({ commentStatus: true });
     } else {
       this.setState({
         showForm: false,
@@ -160,7 +160,7 @@ class Feedback extends React.Component {
                   value={this.state.feedback}
                   onChange={this.handleFeedbackChange}
                 />
-                <div id="textarea-status-note">{this.state.commentStatus}</div>
+                {this.state.commentStatus && <div id="textarea-status-note">Comment Required</div>}
               </div>
 
               <button
--- a/src/app/components/Feedback/Feedback.jsx
+++ b/src/app/components/Feedback/Feedback.jsx
@@ -12,7 +12,6 @@ class Feedback extends React.Component {
     this.state = {
       showForm: false,
       feedback: '',
-      email: '',
       success: null,
     };
 
@@ -23,7 +22,6 @@ class Feedback extends React.Component {
     this.closeForm = this.closeForm.bind(this);
     this.deactivateForm = this.deactivateForm.bind(this);
     this.handleFeedbackChange = this.handleFeedbackChange.bind(this);
-    this.handleEmailChange = this.handleEmailChange.bind(this);
     this.sendFeedback = this.sendFeedback.bind(this);
     this.handleRadioChange = this.handleRadioChange.bind(this);
   }
@@ -73,10 +71,6 @@ class Feedback extends React.Component {
 
   handleFeedbackChange(e) {
     this.setState({ feedback: e.target.value });
-  }
-
-  handleEmailChange(e) {
-    this.setState({ email: e.target.value });
   }
 
   handleRadioChange(e) {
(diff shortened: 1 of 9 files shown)


webpack.config.babel.js
```diff
@@ -21,6 +24,7 @@ const ENV = process.env.NODE_ENV || 'development';
 
 // Sets appEnv so the the header component will point to the search app on either Dev or Prod
 const appEnv = process.env.APP_ENV ? process.env.APP_ENV : 'production';
+const airtableKey = process.env.AIRTABLE_API_KEY || 'broken :(';
```
hokei: lol

hokei:
re: tests - I'd mount it and make sure it has all the right fields, and then do a check to make sure when submit is clicked the right methods get called.  
No biggie though, I think it's just as fair to file a ticket to put it in later, rather than have you trudge through enzyme documentation trying to get tests to work.

src/client/styles/components/Feedback.scss
```diff
@@ -0,0 +1,136 @@
...
+      display: inline-block;
+      width: 0.75rem;
+    }
+
+    .sfr-radio-label {
+      display: inline-block;
+      padding-left: 0.3rem;
+      margin: 0.2rem;
+    }
+  }
```
hokei: nit

hokei:
I can only tab into the "yes" radio button, not the "no"

src/app/components/Feedback/Feedback.jsx
```diff
@@ -0,0 +1,214 @@
...
+    this.openForm = this.openForm.bind(this);
+    this.closeForm = this.closeForm.bind(this);
+    this.deactivateForm = this.deactivateForm.bind(this);
+    this.handleFeedbackChange = this.handleFeedbackChange.bind(this);
+    this.handleEmailChange = this.handleEmailChange.bind(this);
+    this.sendFeedback = this.sendFeedback.bind(this);
+    this.handleRadioChange = this.handleRadioChange.bind(this);
+  }
+
+  onSubmitForm(e) {
```
hokei: Probably needs some sort of error message if you submit the form without the required field
(I'd argue that we probably don't need to require comments, though.  Like, it's useful even to say "Yes I found what I needed")

src/client/styles/components/Feedback.scss
```diff
@@ -0,0 +1,136 @@
+.feedback {
+    position: static;
+    bottom: 0;
+    right: 0;
+    z-index: 100;
+    text-align: right;
+    background: #54514A;
+    color: #000;
+  
+    @include min-screen(1024px) {
```
hokei: nit: I agree it's probably not necessary to have the button in mobile, but is 1024 too big of a min-screen?

mwbenowitz:
> I can only tab into the "yes" radio button, not the "no"

Apparently this is the proper behavior for radio button groups: https://stackoverflow.com/questions/14322564/can-you-tab-through-all-radio-buttons. Which is odd to me but the form does behave as this describes

src/client/styles/components/Feedback.scss
mwbenowitz: This pins the form to the bottom of the window and forces it full width. The original code I borrowed this from has this as a variable "tablet-portrait" and 1024 is the best estimate of that value I could come up with.

In testing it out it seems to work well in Chrome at least.

hokei:
Re: radio buttons -- yeah, it is intended.  It also throws me off every time ... sigh

src/app/components/Feedback/Feedback.jsx
mwbenowitz: Added in the new commits I just pushed!

src/app/components/Feedback/Feedback.jsx
```diff
@@ -0,0 +1,194 @@
...
+                  className="feedback-input"
+                  name="sfr-general-feedback"
+                  rows="5"
+                  aria-required="true"
+                  tabIndex="0"
+                  ref={(textarea) => this.feedbackField = textarea}
+                  value={this.state.feedback}
+                  onChange={this.handleFeedbackChange}
+                />
+                <div id="textarea-status-note">{this.state.commentStatus}</div>
```
hokei: you want `{ this.state.commentStatus && <div id="textarea-status-note">Comment Required</div>}` here
(Conventions - don't want to render this div if you don't have to)

src/app/components/Feedback/Feedback.jsx
```diff
@@ -0,0 +1,194 @@
...
+      </div>
+    );
+  }
+}
+
+Feedback.propTypes = {
+  location: PropTypes.object,
+};
+
+export default Feedback;
```
hokei: nit

PR #111, opened 2020-01-17
Work Detail and Search Results pages

hokei:


hokei:
So this is kind of a monster ... the important things that have changed (and maybe good places to start) are:  
`SearchContainer` -> `LandingPage` and `SearchResultsPage`.  `AdvancedSearchResults` is no longer needed because we use `queries` array for everything now.  
`EditionCard` is a class of static methods that are meant to send stuff to Design System.  The code is finickiest here.  
A lot of supporting elements, such as `Dropdown`, are now moved to Design System and so are removed from here.

src/app/components/WorkDetail/WorkHeader.jsx
```diff
@@ -1,31 +1,34 @@
 import React from 'react';
```
mwbenowitz: This component is throwing an error in my console saying that a list is missing a unique key. I can't see where its coming from though.

src/app/components/WorkDetail/WorkHeader.jsx
```diff
@@ -1,31 +1,34 @@
...
-          <div className="nypl-item-header-author">
-            By&nbsp;
-            <AuthorsList
-              agents={data.agents}
-              max={1}
-              roleFilter="author"
-            />
-          </div>
+const WorkHeader = ({ data }) => {
+  const authorsList = EditionCard.getAuthorsList(EditionCard.getPreferredAgent(data.agents, 'author'));
```
mwbenowitz: This isn't inserting any spaces between the author names.

src/app/constants/editioncard.js
```diff
@@ -0,0 +1,5 @@
+// Constants
+export const MAX_TITLE_LENGTH = 80;
+export const MAX_SUBTITILE_LENGTH = 80;
```
mwbenowitz: Is this coming from the design system? I would think that we could allow the subtitle to be longer than the main title.

src/app/components/Edition/EditionsList.jsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
```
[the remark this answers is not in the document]
hokei: I think it's merge funk - thanks for catching it!

src/app/constants/editioncard.js
hokei: No, we can make it whatever we like, I just picked a number at random haha

mwbenowitz: Ok, cool. Can we make it something like 120? Just thinking that generally the structure (especially for academic books): "Short Title: A Longer Descriptive Title, with stuff" and it would be good to get more of that descriptive sub title in the design if possible

src/app/components/WorkDetail/WorkHeader.jsx
hokei: Yeah I see it too - I think it's from something in the Design System.  I'm not gonna deal with it yet, it seems like a rabbit hole.

src/app/components/Card/EditionCard.jsx
```diff
@@ -0,0 +1,193 @@
...
+    return (
+      `Published ${EditionCard.publisherDisplayLocation(previewEdition)}${EditionCard.publisherDisplayText(previewEdition)}`
+    );
+  }
+
+  // Language Display
+  static getLanguageDisplayText(previewEdition) {
+    let languagesTextList;
+    if (!previewEdition || !previewEdition.languages || !previewEdition.languages.length) {
+      languagesTextList = 'Undetermined';
```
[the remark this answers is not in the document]
hokei: tests caught a few possible bugs so I think it's ... more thorough now!

src/app/constants/editioncard.js
hokei: Yeah!  I have yet to find a book with a subtitle... but that makes lots of sense!

src/app/components/WorkDetail/WorkHeader.jsx
mwbenowitz: For sure, I don't mind the warnings just feel like it should be on a list to address at some point

PR #113, opened 2020-01-21
SFR-442 Add Airtable key to production

mwbenowitz:
To make the Airtable key (loaded as an env var) available to the react app it must be loaded through the webpack config. This is done separately in the production and development environments and the previous commit ommitted the production code, meaning the feedback form cannot be submitted from deployed versions of the app.


hokei:
ah of course.  :shipit:

PR #114, opened 2020-01-27
SFR-593 Fix author name search

mwbenowitz:
The method that constructs author searches form detail pages and search results was passing `name` as the query `field`. This should be `author` the method was returng the same field for both the `query` lookup and the `field` which works for VIAF and LCNAF queries but not for author name searches.

I've updated this to return a tuple, which is then accessed in the calling method. This fixes the issue and should "future-proof" this a bit in case we want to add future searches which may require different field and query names.

Also added are a few tests on the method to ensure that it returns the expected value.

### DIFF
--- a/src/app/components/Card/EditionCard.jsx
+++ b/src/app/components/Card/EditionCard.jsx
@@ -81,12 +81,15 @@ export default class EditionCard {
 
   // Author
   static getAuthorIdentifier(author) {
-    return (author.viaf && 'viaf') || (author.lcnaf && 'lcnaf') || 'name';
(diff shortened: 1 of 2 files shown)


hokei:
Do you think we can remove `query=` entirely?  With this change, I think it isn't (shouldn't be?) used anywhere in the app ...

mwbenowitz:
> edit: hmm this was not using it either, I might be thinking of something different. I'll file a ticket if I stumble across it again

Sounds good. I haven't seen it anywhere, but if it is lingering somewhere we should take it out

PR #117, opened 2020-01-28
SFR-795 read online breadcrumb

hokei:


test/unit/Breadcrumbs.test.js
```diff
@@ -0,0 +1,132 @@
...
+
+import { initialSearchQuery } from '../../src/app/stores/InitialState';
+
+describe('Breadcrumb', () => {
+  const searchQuery = { queries: [{ query: 'snakes', field: 'keyword' }] };
+  const workDetail = { uuid: '12345' };
+  const homeLocation = { pathname: '/' };
+  const otherLocation = { pathName: '/other-location' };
+
+  describe('getBreadcrumbLinks', () => {
```
hokei: would love your thoughts on whether this is too overzealous.  I think some white-box testing is necessary because it doesn't make sense for us to bundle design system and SFR and test them all as a whole, but it might be less brittle to test the contents of the Breadcrumbs render in the L104 block rather than testing these methods separately.  Like, if you were a dev trying to modify breadcrumbs, would these tests be helpful or annoying?

mwbenowitz: I think having both of these tests are helpful. The separate method unit tests give clarity that the methods are performing exactly as anticipated. For me that means that they can help catch edge cases etc. I think that the white-box functional testing is good to verify that the component renders as expected but for me the unit tests are just as important for preventing regressions, the "brittle-ness" might be a good thing since we should be able to very precisely predict those outputs.

src/app/components/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -1,57 +1,97 @@
...
-            {link.text}
-          </Link>
-        )));
-      if (additionalLinks) {
-        crumbs.concat(additionalLinks);
-      }
-    }
-    return crumbs;
-  };
+export const getBreadcrumbLinks = (searchQuery, workDetail) => {
```
mwbenowitz: I like having this method be very explicit about what to output for each page type but I worry about what happens as we add more page types. There aren't that many more (maybe 2?) in the roadmap but if this has to handle any more than that I think this might get unwieldy.

This is definitely a "future-proofing" concern and not something that should block this fix. I think it could be handled pretty easily down the road with an object representing the expected behavior for each page type.

hokei:
I wonder how this is going to evolve as we add more pages ... this code works, but sets off "there must be a better way to do this" bells in my head ... it seems really unweildy for something that is standard for websites ...  I don't have an immediate solution though, I'm sure it'll be revisited as we move on.

PR #118, opened 2020-02-03
Changed to use HTTPS and pointed webpub reader url the friendly one

hokei:


hokei:
oh shoot forgot to update tests.

src/app/util/Util.jsx
```diff
@@ -5,23 +5,17 @@ import FeatureFlags from 'dgx-feature-flags';
...
-  const prefix = 'http://';
+export const formatUrl = (link, env) => {
+  const devPrefix = 'http://';
   const securePrefix = 'https://';
-
-  if (link.substr(0, securePrefix.length) === securePrefix) {
-    // change https to http
-    return prefix + link.substr(securePrefix.length);
+  if (link.substr(0, 4) !== 'http') {
+    return env === 'production' ? securePrefix + link : devPrefix + link;
```
mwbenowitz: If `APP_ENV` isn't set (which should never happen, but you never know) should the app default to `https` or `http`? I tend to think `https` because we want to be sure that stuff works in production and that seems safer to me but I'm not sure.

Either way I think that both this like and the one below should check for either `production` or `development` so that the behavior is consistent if the variable gets mangled somehow.

hokei: I don't think the behavior is consistent either way - in development we have to force `http` even if `https` is passed, but in production we can just pass `http` through (I think we want that).  
I changed it up to throw a warn, I think it's a bit clearer now?

hokei: (Good thing to bring up too, turns out it doesn't get set when you run test ... )

PR #121, opened 2020-02-06
SFR-791 Code cleanup and Unit Tests

hokei:
- Removed a lot of code that was previously merely commented out
- Separated SearchNavigation from SearchPagination
- Separated SearchHeader from SearchForm
- Removed Search Results information from landing page
- Tightened up SearchPagination logic based on tests.


src/app/components/SearchResults/SearchResultsPage.jsx
```diff
@@ -15,6 +13,35 @@ import TotalWorks from '../SearchForm/TotalWorks';
 
 import featureFlagConfig from '../../../../featureFlagConfig';
 import config from '../../../../appConfig';
+import SearchHeader from '../SearchForm/SearchHeader';
+import SearchResults from './SearchResults';
+
+export const loadSearch = (props) => {
```
mwbenowitz: Why move this out of the component? It doesn't seem to be used anywhere else, and I don't think it will be? Not that it doesn't work, but just curious about the move.

src/app/components/SearchResults/SearchResultsPage.jsx
```diff
@@ -116,7 +115,7 @@ class SearchResultsPage extends React.Component {
           >
 
             <div className="sfr-center">
-              <SearchForm
+              <SearchHeader
                 isHomePage={false}
```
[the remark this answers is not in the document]
hokei: That's an oversight, it's not being used anymore!

src/app/components/SearchResults/SearchResultsPage.jsx
hokei: I think I'd moved it out originally to test it in isolation, but it can also be tested using `wrapper.instance()`

src/app/components/WorkDetail/WorkDetail.jsx
```diff
@@ -117,72 +95,76 @@ class WorkDetail extends React.Component {
...
                 {...this.boundActions}
               />
             </div>
           </div>
-          <div className="grid-row">
-            <div className="nypl-item-header">
-              <WorkHeader data={work} />
+          { isValidWork
+          && (
+          <>
```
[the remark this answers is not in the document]
hokei: We used to use `divs` because there wasn't a choice - but ideally, we don't want to nest our code in divs that don't do anything.  Since we're using React 16, might as well make use of this feature!

src/app/components/SearchResults/SearchResultsPage.jsx
mwbenowitz: If it makes testing easier/clearer that's fine by me, just wondering what the reason was

hokei: tbh I'm still waffling on how best to organize the code so that everything can get tested in a sane way... 
so it might get moved back, but for now it doesn't seem like it's causing harm where it is

PR #123, opened 2020-02-10
SFR-811 Consolidated SearchForm and SearchHeader

hokei:
For info on what HOCs are => https://reactjs.org/docs/higher-order-components.html
This is really quite a textbook case to use HOCs, all the lifecycle methods for the two components are the same, but it needs to render different things.  In the past, this could have been done via
```
render() {
return shouldBeSearchPromo ? <SearchPromo /> : <SearchHeader /> 
}
```

(and we did something that was conceptually similar)
but this is way saner

The thing I didn't do was abstract out the unit tests - since they're all on lifecycle and event stuff, it would be a lot of mocking and simulating that I haven't yet figured out how to do without going slightly mad

### DIFF
--- a/src/app/components/SearchForm/WithSearch.jsx
+++ b/src/app/components/SearchForm/WithSearch.jsx
@@ -110,8 +110,8 @@ function withSearch(WrappedComponent) {
 
   SearchComponent.contextTypes = {
     router: PropTypes.objectOf(PropTypes.any),
-    history: PropTypes.objectOf(PropTypes.any),
   };
+
   return SearchComponent;
 }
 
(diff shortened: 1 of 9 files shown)


src/app/components/SearchForm/WithSearch.jsx
```diff
@@ -0,0 +1,118 @@
...
+          Object.assign({}, initialSearchQuery, {
+            query: this.state.searchQuery.queries,
+          }),
+        );
+      }
+    }
+
+    submitSearchRequest(event) {
+      event.preventDefault();
+      const query = this.state.searchQuery.queries[0].query.replace(/^\s+/, '').replace(/\s+$/, '');
```
mwbenowitz: This method for removing leading/trailing whitespace works, but I think `trim()` is shorter and more understandable to anyone who isn't familiar with regexes?

hokei:
Yeah, that's my jam, taking code out of code.  
I think the platonic ideal of HOCs is that they're incredibly re-useable.  I guess this could theoretically be used to put search functionality on everything, but in practice ... we probably don't want search functionality on everything haha

PR #124, opened 2020-02-10
SFR-804 invalid Search url routing

hokei:
Changes unintended search URLs to redirect to landing page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20069,6 +20069,12 @@
         }
       }
     },
+    "sinon-chai": {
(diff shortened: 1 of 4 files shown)


hokei:
Do you know off the top of your head if a result like this will appear in GA?  It'd be nice to know if there's someone whose consistently trying to fudge with urls haha

mwbenowitz:
Yeah, we track all traffic on the site, so we would see the weird url show up and be able to tell if people are messing with them

PR #125, opened 2020-02-18
SFR-89 request digitization

hokei:
Added a Request Digitization form based on Feedback. 
Incorporated this form into WorkDetail and ResultsList
Made it so feature flag can be changed via the URL
Made changes to the readme to reflect this


hokei:
I feel like I was able to do this in the past, but googling is coming up short -- The tests are failing because it can't install the design system from a branch.  It seems to be a Lerna issue.  For now the only way to run it is to check out the branches and link them.

src/app/components/List/EditionsList.jsx
```diff
@@ -3,21 +3,33 @@ import PropTypes from 'prop-types';
...
   componentDidMount() {
     this.setState({ loaded: true });
   }
 
+  getAllEditionsData(work, editions, origin, eReaderUrl, referrer) {
+    /* eslint-disable-next-line no-underscore-dangle */
+    return editions.map(
+      (edition) => {
+        const showRequestButton = this.props.getRequestEditionButton(edition);
+        const blah = EditionCard.getEditionData(edition, origin, eReaderUrl, referrer, showRequestButton);
```
mwbenowitz: Can probably just return this directly (or replace this excellent variable name haha)

hokei:
re: doing this via postinstall 
I came across this thing - https://stackoverflow.com/questions/39194648/npm-install-package-from-github-repo-subfolder which seems like a loooot more effort than it's worth.  I think if we really need it, it might make more sense to publish to npm as `-alpha` and use that instead ...

mwbenowitz:
haha yeah that seems way overcomplicated to simply clone a branch of a git repo, I did it with `git clone --single-branch --branch SFR-89_request_digitization_fake https://github.com/NYPL/nypl-design-system.git`, but when I install from that it doesn't recognize that the package is installed. I'll think about it for a little bit more but if I/we can't figure it out in the next hour or two I say we just merge it and do a manual deploy for this until the design system gets merged and released

hokei:
"For the feature flag, is there a way to make it clear if its activated?" 
Right now, if the url has `feature=` and it's deactivated if you remove it.  It doesn't read the `featureflagconfig` object and there's no testing on what happens if you are using both at the same time.  
Right now I think consider the URL stuff a hack - to do it properly, we will want to move the `dgx-feature-flags` package to github and and add this functionality there.

mwbenowitz:
oh okay, I misunderstood the code then!

I'm going to approve this. It seems silly to spend too much time worrying about a workaround for the design system stuff when it will be updated shortly. If it keeps on coming up we should try to find a solution, but that's a question for another day.

PR #130, opened 2020-02-21
SFR-135 license explanations

hokei:
[Jira Ticket](http://jira.nypl.org/browse/SFR-135)

- Added License Explanations page and links to it from the edition card

-  Run SFR and make sure that license links to the "License Explanations" page.  
- Will there always be only one license?  This model breaks if there's more than one license ...

### DIFF
--- a/src/app/components/License/License.jsx
+++ b/src/app/components/License/License.jsx
@@ -1,0 +1,62 @@
+import React from 'react';
+import { Html5Entities } from 'html-entities';
+import * as DS from '@nypl/design-system-react-components';
+import { Link } from 'react-router';
+import { detailDefinitionLabels } from '../../constants/labels';
+import {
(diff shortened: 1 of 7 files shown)


src/app/components/License/License.jsx
```diff
@@ -0,0 +1,62 @@
...
+import { Html5Entities } from 'html-entities';
+import * as DS from '@nypl/design-system-react-components';
+import { Link } from 'react-router';
+import { detailDefinitionLabels } from '../../constants/labels';
+import {
+    unique, flattenDeep, isEmpty, uniqueAndSortByFrequency,
+} from '../../util/Util';
+
+export const License = () => {
+    return (<DS.Container>
```
mwbenowitz: No breadcrumb! Without it there is no obvious link back to ResearchNow

hokei: d'oh!  Yes I will add this!

PR #131, opened 2020-02-21
SFR-808 verify ds components

hokei:

- Increases test coverage by adding tests for EditionList and expanding tests for WorkDetail
- After this PR, SearchPagination and SearchNavigation should be the only major things that main untested (TotalWorks is also untested, but much lower priority) 

-  Make sure things don't break ... ?

The commit history looks really ugly and I'm not really sure how to squash stuff when they're interspersed with all these merge commits.


test/unit/SearchPagination.test.js
```diff
@@ -9,6 +9,15 @@ import SearchPagination from '../../src/app/components/SearchResults/SearchPagin
...
+  describe('Get Page List', () => {
+    it('returns list of pages', () => {});
+  });
+
+  describe('calling navigateToPage changes query and calls submit', () => {
+  });
+
+  describe('Changing select calls changes searchQuery', () => {});
+
   const query = { query: 'Chicago', field: 'keyword' };
```
mwbenowitz: I know this isn't part of the commit, but should it be wrapped in a `before` call?

PR #132, opened 2020-02-24
SFR-89 request digitization

hokei:
[Jira Ticket](http://jira.nypl.org/browse/SFR-89)

- Fixes some bugs that come out of QA: 
-- Modal appears on top of fotoer
-- Request digitization requires checking a radio button
-- Small copy tweak

-  Verify the above

### DIFF
--- a/src/app/components/SearchResults/ResultsList.jsx
+++ b/src/app/components/SearchResults/ResultsList.jsx
@@ -68,7 +68,15 @@ class ResultsList extends React.Component {
 
     return results.map((result, index) => {
       const showRequestButton = shouldShowRequest ? (
-        <a class="link" onClick={() => this.openForm(result, result.editions[0])}>Request Digitization</a>
+        <a
+          role="link"
+          tabIndex="0"
+          className="link"
+          onKeyDown={() => this.openForm(result, result.editions[0])}
+          onClick={() => this.openForm(result, result.editions[0])}
+        >
+        Request Digitization
+        </a>
       ) : undefined;
 
       const titleElement = EditionCard.generateTitleLinkElem(result.title, result.uuid);
--- a/src/app/components/SearchResults/ResultsList.jsx
+++ b/src/app/components/SearchResults/ResultsList.jsx
@@ -72,7 +72,7 @@ class ResultsList extends React.Component {
           role="link"
           tabIndex="0"
           className="link"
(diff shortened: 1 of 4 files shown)


src/client/styles/components/RequestDigitization.scss
```diff
@@ -18,6 +18,7 @@
     box-shadow: -1px 0 1px 0 rgba(0,0,0,0.3);
     padding: 0.35rem 0.5rem;
     text-align: left;
+    z-index: 100;
```
mwbenowitz: Is this just for safety or is there a known case that can cause something else to appear in front of the form?

src/app/components/SearchResults/ResultsList.jsx
```diff
@@ -68,7 +68,15 @@ class ResultsList extends React.Component {
 
     return results.map((result, index) => {
       const showRequestButton = shouldShowRequest ? (
-        <a class="link" onClick={() => this.openForm(result, result.editions[0])}>Request Digitization</a>
+        <a
+          role="link"
+          tabIndex="0"
+          className="link"
+          onKeyDown={() => this.openForm(result, result.editions[0])}
```
mwbenowitz: Should this be tied explicitly to the enter key? My event knowledge is a bit rusty but I think this would trigger the form on any key press?

src/client/styles/components/RequestDigitization.scss
hokei: Yeah, Chris found it:  
https://jira.nypl.org/browse/SFR-89

It comes from the desire to do this in as quick-and-dirty of a way as possible ... a more proper way to do this would be to place it in the center of the screen with all the correct screen positioning calculations.  

The hope is that when this gets implemented properly (not as a modal), it would be using a different class and this class will be eliminated, because z-index like this is ... generally a hack.

src/app/components/SearchResults/ResultsList.jsx
hokei: I was supressing a warning without thinking about it -- you're right, we should submit on enter

src/client/styles/components/RequestDigitization.scss
mwbenowitz: haha yeah, z-index stuff can definitely get out of hand, we aren't using too many modals though and this definitely works to address the issue for now

PR #135, opened 2020-02-27
SFR-817 removed search results from breadcrumb and added work names t…

hokei:
[Jira Ticket](http://jira.nypl.org/browse/SFR-817)

- Changes the Read Online links to use router
- Removes passing of SearchQuery into breadcrumb
- Passes work from read online page to breadcrumb

-  Breadcrumbs show up as needed

### DIFF
--- a/src/app/components/List/EditionsList.jsx
+++ b/src/app/components/List/EditionsList.jsx
@@ -10,20 +10,13 @@ class EditionsList extends React.Component {
     this.props = props;
 
     this.getAllEditionsData = this.getAllEditionsData.bind(this);
-    this.state = {
-      loaded: false,
-    };
   }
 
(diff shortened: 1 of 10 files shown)


src/app/components/List/EditionsList.jsx
```diff
@@ -10,20 +10,13 @@ class EditionsList extends React.Component {
...
-    };
   }
 
-  componentDidMount() {
-    this.setState({ loaded: true });
-  }
-
-  getAllEditionsData(editions, origin, eReaderUrl, referrer) {
-    return editions.map(
+  getAllEditionsData(work, origin, eReaderUrl, referrer) {
```
mwbenowitz: `origin` isn't used here and isn't in the invocation below

PR #137, opened 2020-03-03
NO-REF releasing 8.2

hokei:
No Jira Ticket

- Bumps package and changelog for release

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 # CHANGE LOG
 
-## PRERELEASE 
+## [0.8.2] 
 ### Added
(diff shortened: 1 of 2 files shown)


hokei:
shouldn't this be pushed automatically with tag?  Is there something else we need to set up that isn't set up?

hokei:
(I think maybe I forgot to commit before pushing the tag)

mwbenowitz:
Yeah, you have to bump the version manually, I don't know of any integration that can do it automatically

hokei:
Yeah I think what I did was deploy without these bumps ... will remember for next time
or bring up a checklist haha

PR #148, opened 2020-03-20
SFR-843 adv search persistent state

hokei:
[Jira Ticket](http://jira.nypl.org/browse/SFR-843)

- Prepopulates advanced search fields based on visible queries (This is more clear than populating on non-visible queries in unlikely case of conflict) 
- Clears non-visible queries when moving to adv search.

-

### DIFF
--- a/src/app/components/AdvancedSearch/AdvancedSearch.jsx
+++ b/src/app/components/AdvancedSearch/AdvancedSearch.jsx
@@ -108,6 +108,11 @@ class AdvancedSearch extends React.Component {
     this.setState((prevState) => {
       const queries = prevState.queries;
       queries[name] = value;
+      if (name === 'author') {
+        // Changing author should clear viaf and lcnaf
+        queries.viaf = '';
+        queries.lcnaf = '';
+      }
       return { queries, error: false, errorMsg: '' };
     });
   }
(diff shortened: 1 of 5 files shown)


src/app/components/AdvancedSearch/AdvancedSearch.jsx
```diff
@@ -215,9 +208,21 @@ class AdvancedSearch extends React.Component {
...
+    if (searchQuery.showField && searchQuery.showQuery) {
+      state.queries[searchQuery.showField] = searchQuery.showQuery;
+    }
+
     if (searchQuery.queries) {
       searchQuery.queries.forEach((q) => {
-        state.queries[q.field] = q.query;
+        const allowedFields = inputTerms.map(term => term.values.map(value => value.key)).flat();
+        if (allowedFields.indexOf(q.field) >= 0) {
+          // If field is already set (by showQuery), use the showQuery value.
```
mwbenowitz: If the show query is an author name, I think this should still query with the VIAF identifier? I'm not 100% clear on what this is doing though, so maybe that's already addressed. Otherwise this is a safe assumption to make.

hokei: You're right:  Right now if someone clicks the author name, it looks it up by viaf, but if they hit advanced search and then hit search again, it will only search by name.  

Is there a way to look up the author/publisher from the viaf?  I'm worried about a situation where someone clicks an author name, so it searches by viaf, then they change the name in advanced search, but the viaf (for the original author) still persists.  How do we check that the viaf still pertains to the author?

mwbenowitz: hmm, good point, there is an endpoint to look up VIAF identifiers, but I'm not sure we should be using that from the front end...

could you do it from the `onQueryChange` method? Check if the field is author and then wipe the viaf and lcnaf fields in state?

hokei: Oh that's a great idea!  Let's do that!

PR #149, opened 2020-03-24
NO-REF re-implemented landing promo so that the header displays corre…

hokei:
…ctly

- Until the hero re-factor work is done, let's just use a local version of the header so that it's not causing highly visible weird-ness 

-

### DIFF
--- a/src/app/components/LandingPage/LandingPage.jsx
+++ b/src/app/components/LandingPage/LandingPage.jsx
@@ -61,16 +61,33 @@ class LandingPage extends React.Component {
           />
           <div
             aria-label="Digital Research Books Beta"
-          >
-            <DS.HeaderImgRight
-              headerId="ResearchNow-Main-Header"
-              isImageDecorative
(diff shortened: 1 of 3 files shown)


src/app/components/LandingPage/LandingPage.jsx
```diff
@@ -61,16 +61,33 @@ class LandingPage extends React.Component {
...
+                  Find millions of digital books for research
+                  from multiple sources world-wide--all free to read, download, and keep. No library card required.
+                  This is an early beta test, so we want your feedback!
+                  {' '}
+                  <DS.UnderlineLink><Link to="/about">Read more about the project</Link></DS.UnderlineLink>
+                  .
+                </p>
+              </div>
+            {/* TODO: Use DS again after Hero is re-written. 
+              <DS.HeaderImgRight
```
scooooooooby: If you want to get ahead of it here, the refactor renames this to `Hero`

PR #150, opened 2020-03-25
SFR-843 Separated out Query and ShowQuery 

hokei:
[Jira Ticket](http://jira.nypl.org/browse/SFR-843)

- Query and ShowQuery were inferring values off of each other.  
- This PR separates them so that they are always set differently.  This allows us to handle queries that are used in search (eg: viaf) that are different from what is shown to the user,  (eg: if searching by viaf, should override author search, but author search should be used when no viaf is available) 

-

### DIFF
--- a/src/app/components/AdvancedSearch/AdvancedSearch.jsx
+++ b/src/app/components/AdvancedSearch/AdvancedSearch.jsx
@@ -19,7 +19,7 @@ import FilterYears from '../SearchResults/FilterYears';
 import { searchFields } from '../../constants/fields';
 
 
-const initialState = {
+export const initialState = {
   error: false,
   errorMsg: '',
   languages: [],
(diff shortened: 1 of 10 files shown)


src/app/components/AdvancedSearch/AdvancedSearch.jsx
```diff
@@ -225,13 +241,13 @@ class AdvancedSearch extends React.Component {
...
     }
-    // If viaf or lcnaf is set, but author is not set, they were publisher searches
-    // Since we don't give users a way to clear this, we should
-    // pre-emptively clear it.
-    if (!state.queries.author) {
-      state.queries.viaf = '';
-      state.queries.lcnaf = '';
-    }
+
+    const displaySearchQueries = searchQuery.queries.filter(query => searchFields.includes(query.field));
```
mwbenowitz: Won't this preserve `lcnaf` and `viaf` fields in the `displayQueries` object? Or am I missing something?

hokei: No `filter` checks for whether the field exists in `searchFields` ( https://github.com/NYPL/sfr-bookfinder-front-end/pull/150/files#diff-0565b36c278253c00836552fe5428f75R9 )

mwbenowitz: ah ok! I got `searchFields` mixed up with `fields`

PR #155, opened 2020-04-02
SFR-893 Add Production Step to TravisCI

mwbenowitz:
[Jira Ticket](http://jira.nypl.org/browse/SFR-893)

Adds a production deployment step to the TravisCI build pipeline. It should automatically deploy to the production instance when changes are merged into the `beta-production` branch.

No testing requirements for this. Due to the nature of TravisCI it can only be tested when merged into the appropriate branch


hokei:
will travis bump package.json automatically?

mwbenowitz:
It won't, just an oversight on my part. So bumping that here, then will merge and open a PR to `beta-production` that we can merge on Monday

PR #157, opened 2020-04-03
HOTFIX Remove HTTP protocol coercion

mwbenowitz:
HOTFIX - No JIRA ticket

The `formatUrl` shim was put in to force URLs to use a certain protocol for environmental reasons (generally being forced to https). Due to updates in the environments this is no longer necessary.

The method is still in place but now simply adds a protocol to any URL that is missing one (this is necessary, for example, with covers). This should allow for certain components, such as the reader app, to work as intended in all environments.

This also adds a new environment: `local`. This is similar to `development` but has some specific settings to allow for easier local development. It can be started by using `npm run local-start` and requires that a local instance of the webpub-reader app be running at `localhost:4444`

- Run a local instance of the app with `npm run local-start`
- Verify that all resources load properly with a focus on covers, the ePub reader and other 3rd party readers embedded via iFrame

### DIFF
--- a/src/app/components/Card/EditionCard.jsx
+++ b/src/app/components/Card/EditionCard.jsx
@@ -170,7 +170,7 @@ export default class EditionCard {
   static generateStreamedReaderUrl(url, eReaderUrl, referrer) {
     const base64BookUrl = Buffer.from(formatUrl(url)).toString('base64');
     const encodedBookUrl = encodeURIComponent(`${base64BookUrl}`);
-    console.log(eReaderUrl)
+
     let combined = `${eReaderUrl}/readerNYPL/?url=${eReaderUrl}/pub/${encodedBookUrl}/manifest.json`;
     if (referrer) {
       combined += `#${referrer}`;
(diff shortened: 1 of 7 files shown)


src/app/components/Card/EditionCard.jsx
```diff
@@ -168,9 +168,9 @@ export default class EditionCard {
   // This is specific to and backwards-engineered from the webpub-viewer URLs.
   // and should be changed when webpub-viewer is able to generate more reasonable URLs.
   static generateStreamedReaderUrl(url, eReaderUrl, referrer) {
-    const base64BookUrl = Buffer.from(formatUrl(url, process.env.APP_ENV)).toString('base64');
+    const base64BookUrl = Buffer.from(formatUrl(url)).toString('base64');
     const encodedBookUrl = encodeURIComponent(`${base64BookUrl}`);
-
+    console.log(eReaderUrl)
```
hokei, line 173: console.log

mwbenowitz: Thanks!

hokei:
is the idea that running `local` should point you to the local copy of webpub-viewer?  (development still points to the local copy)

mwbenowitz:
Yeah, I think I fixed the `development` / `local` issue in webpack in the last push I just made. But yes, they should be the same with the exception of which webpub-viewer is being used. I also removed the GA tracking code from local so at least local test sessions won't show up as users.

PR #158, opened 2020-04-06
SFR-885 Add GA tracking events for Read Online and Download buttons

mwbenowitz:
[Jira Ticket](http://jira.nypl.org/browse/SFR-885)

- Adds `onClick` events to the "Read Online" and "Download" buttons in edition cards that trigger GA tracking events that record which button was clicked, the source of the edition being accessed and the title of the work

-  Verify that the onClick triggers are present in all cases
- (Optional) trigger an event and verify that it appears in the Google Analytics reports

### DIFF
--- a/src/app/components/Card/EditionCard.jsx
+++ b/src/app/components/Card/EditionCard.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { Link } from 'react-router';
+import { gaUtils } from 'dgx-react-ga';
 import { Html5Entities } from 'html-entities';
 import * as DS from '@nypl/design-system-react-components';
 import {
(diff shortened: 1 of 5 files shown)


src/app/components/Card/EditionCard.jsx
```diff
@@ -198,16 +200,30 @@ export default class EditionCard {
...
+          className="edition-card__card-info-link"
+          href={`${formatUrl(selectedLink.url, process.env.APP_ENV)}`}
+          onClick={() => gaUtils.trackGeneralEvent('Download', editionItem.source, work.title, '')}
+        >
+          Download
+        </a>
+      );
+    }
+
+    return undefined;
```
hokei, line 226: this shouldn't be reachable

hokei:
Verified that the events with labels show up on GA, but we probably need to deploy to have new triggers show up?  (My local work clicking doesn't seem to be adding new events yet)

hokei:
:shipit:

src/app/components/Card/EditionCard.jsx
mwbenowitz: Just removed it in the merge to update the `formatUrl` calls

PR #162, opened 2020-04-21
SFR-640 definition list

hokei:

- Changes the Definition List to use `<dl>` and children elements
- Cleaned up styles for definition list

-


src/app/constants/labels.js
```diff
@@ -1,18 +1,10 @@
 export const detailDefinitionLabels = {
-  // also provides sort for DefinitionList
   alt_titles: 'Alternative Titles',
-  medium: 'Medium',
-  series: 'Series',
```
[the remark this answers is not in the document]
hokei: this definitely was my mistake! good catch!

src/app/constants/labels.js
```diff
@@ -1,18 +1,10 @@
...
   summary: 'Summary',
+  series: 'Series',
```
hokei, line 4: @mwbenowitz haha I thought I had been seeing series - it's still here!

mwbenowitz: whoops haha

PR #165, opened 2020-04-30
SFR-902 edition detail

hokei:

- Adds Edition Detail page

-  Make sure edition detail loads and functions as expected


src/app/components/Breadcrumbs/Breadcrumbs.jsx
```diff
@@ -1,21 +1,36 @@
...
+import EditionCard from '../Card/EditionCard';
 
-export const getBreadcrumbLinks = (workDetail) => {
+const getLinkFromWork = (title, uuid) => {
+  const workTitle = title || 'Title Unknown';
+
+  const strippedSlashTitle = workTitle.indexOf('/') > 0
+    ? workTitle.substring(0, workTitle.indexOf('/')) : workTitle;
+  const breadcrumbTitle = strippedSlashTitle.length > MAX_TITLE_LENGTH
+    ? `${strippedSlashTitle.substring(0, MAX_TITLE_LENGTH)}...` : strippedSlashTitle;
```
mwbenowitz, line 16: Wonder if it makes sense to break the title here on a word boundary as well

src/app/components/Card/EditionCard.jsx
```diff
@@ -55,26 +52,31 @@ export default class EditionCard {
...
       </Link>
     );
   }
 
+  static generateDisplayTitle(work) {
+    let displayTitle;
+    if (!work.title) {
+      displayTitle = 'Title Unknown';
+    } else if (work.title.length > MAX_TITLE_LENGTH) {
+      displayTitle = `${work.title.substring(0, MAX_TITLE_LENGTH)}...`;
```
mwbenowitz, line 73: Same question on title breaking consistency

src/app/components/Card/EditionCard.jsx
```diff
@@ -258,6 +270,23 @@ export default class EditionCard {
     );
   }
 
+  static getWorldCatElem(instance) {
+    const oclc = instance && instance.identifiers
+      ? instance.identifiers.find(identifier => identifier.id_type === 'oclc').identifier : undefined;
+    const oclcLink = oclc ? `http://www.worldcat.org/oclc/${oclc}` : undefined;
```
mwbenowitz, line 276: Minor but this should be `https`

src/app/components/Breadcrumbs/Breadcrumbs.jsx
hokei: the word boundary ticket hasn't been done yet, but I will make a note to address breadcrumbs in that ticket!

src/app/components/Detail/WorkDetail.jsx
```diff
@@ -45,9 +44,10 @@ class WorkDetail extends React.Component {
 
   componentDidMount() {
     const { query, hash } = this.props.location;
-    const workId = query && query.workId;
     this.setState({ shouldShowAll: query && query.showAll ? (query.showAll === 'true') : true });
-    this.loadWork(workId, hash, this.boundActions);
+    if (query) {
```
mwbenowitz, line 48: If `query` is missing should an error be thrown, or a redirect happen back to the landing page?

src/app/components/Viewer/EBookViewer.jsx
```diff
@@ -40,15 +40,16 @@ class EBookViewer extends React.Component {
...
+    const edition = this.props.location.state ? this.props.location.state.edition : undefined;
     const bookUrl = this.state.bookUrl;
 
     return (
       <span>
         <Breadcrumbs
           router={router}
           location={this.props.location}
-          workDetail={work}
+          workDetail={{ uuid: edition.work_uuid, title: edition.title }}
```
mwbenowitz, line 51: This is just a note, edition titles and work titles can be different. We try to choose the most popular or original edition title for the work, but they can be different. It's usually not noticeable but it might be something that we want to correct in the future.

test/unit/SearchActions.test.js
```diff
@@ -37,4 +37,19 @@ describe('SearchActions', () => {
       expect(JSON.stringify(workDetailAction.work)).to.equal(JSON.stringify(details));
     });
   });
+
+  describe('editionDetail Action', () => {
+    it('should return an object with type and editionDetail', () => {
```
mwbenowitz, line 42: Should this test be commented out?

test/unit/EditionDetail.test.js
```diff
@@ -0,0 +1,67 @@
...
+    it('should show ResultsHeader', () => {
+      expect(container.find('SearchComponent').dive().find('ResultsHeader').exists()).to.equal(true);
+    });
+
+    it('should show work header', () => {
+      expect(container.find('#edition-title').exists()).to.equal(true);
+    });
+
+    it('should show a EditionDetailDefinitionList', () => {
+      expect(container.find('EditionDetailDefinitionList').exists()).to.equal(true);
```
mwbenowitz, line 58: I think this should verify that a few bits of the metadata from the sample edition appear where we expect them to, like languages and publishers maybe. Just to verify that it seems like its loading data properly.

src/app/components/Viewer/EBookViewer.jsx
hokei: Yeah it's ... tricky.  I did it this way to try not to pass in work into state (we don't always have work in state, (clicking "read online" from the Edition Detail page) but we also don't always have edition.title in state (clicking "read online" from the Search Results page).  It makes it possible that, depending on where you clicked "Read Online" from, you might get a different work name.   (Just like if you navigate to a reader page from a link someone sends you, you won't get work and edition in breadcrumbs)  This ... can't be helped, unless we have the "Read Online" page be an "Item" page where "Item" knows what work and edition belongs to.

mwbenowitz: yeah, I think it's fine for it to work this way. It's actually even more complicated because the item you're viewing might actually have a different title from the work, so then the question becomes: which is the right title to display? So in short: definitely should just be left as is, I'm just trying to flag stuff like this so we know its out there.

hokei: sounds good.  Books, huh.

test/unit/EditionDetail.test.js
hokei: this is done in `EditionDetailDefinitionList.test.js` -- we shouldn't reach into a child component here.

src/app/components/Viewer/EBookViewer.jsx
hokei: OK so right now the breadcrumbs will always use the work title that's in edition if there is one, but it will substitute with the work title if there isn't one.  That's happening in `EditionCard.getReadOnlineLInk`

src/app/components/Detail/EditionDetail.jsx
```diff
@@ -0,0 +1,267 @@
...
+      shouldShowAll: true, ...props,
+    };
+    this.toggleReadOnline = this.toggleReadOnline.bind(this);
+    this.boundActions = bindActionCreators(searchActions, dispatch);
+  }
+
+  componentDidMount() {
+    const { query, hash } = this.props.location;
+    this.setState({ shouldShowAll: query && query.showAll ? (query.showAll === 'true') : true });
+    console.log('got here');
```
mwbenowitz, line 44: Noticed a few stray `console.log` statements

PR #168, opened 2020-05-06
SFR-902 edition detail bugfixes

hokei:
[Jira Ticket](http://jira.nypl.org/browse/SFR-902)

- Bugfixes for the "Show All" toggle and style fixes for definitionlists

-


src/app/components/Detail/WorkDetailDefinitionList.jsx
```diff
@@ -151,7 +155,7 @@ export const WorkDetailDefinitionList = ({ work }) => {
           (entry, i) => elements.includes(entry[0]) && entry[1] && (
           <React.Fragment key={`"entry"${i.toString()}`}>
             <dt>{workDetailDefinitionLabels[entry[0]]}</dt>
-            <dd>{parseEntries(entry[0], entry[1], workObj)}</dd>
+            { parseEntries(entry[0], entry[1], workObj) && <dd>{parseEntries(entry[0], entry[1], workObj)}</dd> }
```
mwbenowitz, line 158: What am I missing here? `parseEntries` always returns an element, so why is the check needed?

src/app/components/Detail/WorkDetail.jsx
```diff
@@ -140,6 +140,10 @@ class WorkDetail extends React.Component {
     } else if (query.showAll) {
       query.showAll = 'true';
       this.setState({ shouldShowAll: true });
+    } else {
+      // No showAll in query
+      query.showAll = 'false';
+      this.setState({ shouldShowAll: false });
```
mwbenowitz, line 146: should `false` be a string here? seems inconsistent with the change above

src/app/components/Detail/WorkDetailDefinitionList.jsx
hokei: Aaaahh hmmm I was under the impression they didn't but you're right that they should - this shouldn't get here if parseEntries isn't doing anything.  I'll do a double-check.

mwbenowitz: I think(?) the checks on 155 handle it? It seems like it tests if the field exists on the record and if the value is present before returning.

It certainly doesn't hurt anything, just trying to avoid doubling the number of times that method gets called.

src/app/components/Detail/WorkDetail.jsx
hokei: It should be a boolean - query.showAll is a string because that's is expression in the query url string, but it's much simpler to set the state to be boolean values.  

Or, put in a robust "parseQueryToString" and "parseStringToQuery", but since showAll is the only thing that would need to be transformed, it might be overkill.

hokei: Oh also, this is what fixes the subject bug - before, if you didn't pass a showAll into query, it would error.  The subject links don't pass showAll into query (they could, but this is still a case that should be handled)

PR #169, opened 2020-05-07
SFR-789 Break Strings on Whitespace

mwbenowitz:
[Jira Ticket](http://jira.nypl.org/browse/SFR-789)

- Adds a utility function that truncates strings on word boundaries

- Verify that titles and sub-titles are properly displayed, with ellipsis appearing only at the end of words, not in the middle of them

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,9 @@
 # CHANGE LOG
-## [0.9.2]
-### Added
-- String truncating now occurs on word boundaries
 
 ## [0.9.1] 
 - Updated dgx-global-footer to 0.5.5.
 - Edited some typos and added page title to Edition Detail
+- String truncating now occurs on word boundaries
 
 ## [0.9.0] 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 # CHANGE LOG
+## [0.9.2]
+### Added
+- String truncating now occurs on word boundaries
+
 ## [0.9.1] 
 - Updated dgx-global-footer to 0.5.5.
 - Edited some typos and added page title to Edition Detail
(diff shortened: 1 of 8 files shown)


src/app/util/Util.jsx
```diff
@@ -79,6 +79,20 @@ export const checkFeatureFlagActivated = (featureFlagList, componentStateObject)
...
+ * truncateStringOnWhitespace(str, maxLength)
+ * Return a version of the string shortened to the provided maxLength param. This includes
+ * the three characters for the ellipsis that is appended. If the string is shorter than the
+ * max length it is returned as is.
+ * @param {string} str - The string to be shortened (or returned without change)
+ * @param {int} maxLength - The maximum length of the returned string to be applied
+ */
+export const truncateStringOnWhitespace = (str, maxLength) => {
+  if (str.length < maxLength) { return str; }
+  const truncArray = str.substr(0, maxLength - 3).split(/\s+/).slice(0, -1);
```
hokei, line 92: There is an edge case here that seems incredibly rare, but ... what if the entire title is one long word?  (input errors, may be?). In this case the title would be "..." because I think slice would return "".  I think that's ok - the other way to do it would be to truncate not on whitespace.

hokei: either way, let's add a test for this so that it's documented.

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # CHANGE LOG
+## [0.9.2]
```
hokei, line 2: So we won't have released 9.1 (or 9.0) before this is in, should we be adding to the change log instead of skipping versions?

src/app/util/Util.jsx
mwbenowitz: I think in that instance we should truncate not on whitespace, otherwise it will look to the user that we don't have a title, which is worse than ugly truncation I think.

But will definitely add a test

CHANGELOG.md
mwbenowitz: oh yeah, that's just an oversight on my part then, I'll switch this back to 9.1

PR #171, opened 2020-05-13
SFR-821 persistent searchterm

hokei:
[Jira Ticket](http://jira.nypl.org/browse/SFR-821)

- When you search from the landing page, the search term also persists in the Search Results page.  
- As a consequence of this, if you search with one field in the advanced search page, it will also display in the searchbar.

-


src/app/components/SearchForm/WithSearch.jsx
```diff
@@ -25,6 +27,15 @@ function withSearch(WrappedComponent) {
 
     componentDidMount() {
       global.window.scrollTo(0, 0);
+      this.setState({ searchQuery: this.props.initialQuery });
+    }
+
+    componentDidUpdate(prevProps) {
+      // Wait for initial query to be returned before setting it in state
+      if (this.props.initialQuery && this.props.initialQuery !== prevProps.initialQuery) {
```
hokei, line 35: This seems dumb and unnecessary, but it's not working without it.  Tagging @EdwinGuzman for a second opinion.

PR #172, opened 2020-05-18
SFR-852 Add Citation Parser

mwbenowitz:
[Jira Ticket](http://jira.nypl.org/browse/SFR-852)

### Questions I have
- How can we style the citations simply so that they appear visually "attached" to the cards above them?
- Does the wrapper function make sense?
- Did I implement the feature flags properly?
- Did I structure the output of the `List` elements properly?
- How will we address the need to implement this on the edition page?

- Adds an APA citation below each edition card
- Places this feature under a `DisplayCitations` flag

-  [Run unidentified/unwritten tests]


src/app/components/Citations/formatCitation.jsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+function withCitationFields(WrappedComponent) {
```
hokei, line 5: At a high level, this class seems to be taking props.edition and transforming it into something where you can get Title, etc.  

I think this is better suited to be a utility class like `EditionCard.jsx` where it just exposes a bunch of functions like (getAuthor), (getTitle), etc.  There might be overlap with `EditionCard`? Though I think it's not a bad thing to keep the two separate even if they happen to be returning the same results right now.

src/app/components/List/EditionsList.jsx
```diff
@@ -1,35 +1,68 @@
...
 
   getEditionCardList(work, eReaderUrl, referrer) {
     return work.editions.map(
       (edition, index) => {
         const showRequestButton = this.props.getRequestEditionButton(edition);
         const editionData = EditionCard.getEditionData(work, edition, eReaderUrl, referrer, showRequestButton);
+
+        const shouldShowCitations = FeatureFlags.store._isFeatureActive(config.displayCitations.experimentName);
+        const citationElement = shouldShowCitations ? (
+          <FormattedAPACitation
```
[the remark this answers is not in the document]
hokei: Here I'd say something like parsing out initials is part of "display", even though /technically/ it might not be.  

Also, do we expect last names and first names to always be well labeled?

src/app/components/Citations/APACitation.jsx
```diff
@@ -0,0 +1,170 @@
...
+        initials = getInitials(initialArray);
+      }
+      if (initFirst) {
+        return `${initials} ${lastName}`;
+      }
+      return `${lastName}, ${initials}`;
+    }).join(', ');
+  }
+
+  render() {
```
hokei, line 110: This is definitely stylistic, but I am finding this file pretty hard to read.  

I think partly, it's because there's both logic and presentation here, and I think separating it out (see comment in `EditionsList`).  But as a general rule of thumb, since this is mostly a presentation element, I would aim to have most of the markup live here rather than in separate functions.  In situations where you want markup to live in separate functions, I wouldn't nest them more than one level 

(eg: OK, but still better avoided, to go, `<div>{this.getElement()}</div>` but try to avoid `<div>{this.callAnotherFunctionThatReturnsAnElement()}</div>`) 

So here, I think it best to try to have it go:  

```
return (
      <div className="apa-citation body-small">
        <strong>APA</strong>
        <br />
        {isGovernmentReport && returnGovernmentReport()}
        {!isGovernmentReport && returnMonographCitation(isVolume)}  
      </div>
)
```

src/app/components/Citations/APACitation.jsx
```diff
@@ -0,0 +1,170 @@
...
+
+class APACitation extends React.Component {
+  componentDidMount() {
+    this.getComponentStrs();
+  }
+
+  getComponentStrs() {
+    this.authorsStr = this.formatAgentNames('authors');
+    // eslint-disable-next-line max-len
+    this.editorStr = `${this.formatAgentNames('editors', this.props.agents.authors.length > 0)}, (Ed${this.props.agents.editors.length === 1 ? '.' : 's.'})`;
```
hokei, line 15: js stylistic thing, there's no need to postfix everything with Str, 

perhaps "formattedAuthors" to imply that stuff has been already transformed?  

Also, it's pretty useful to know if stuff is an array or not by the presence/absence of the pluralizer

src/app/components/List/EditionsList.jsx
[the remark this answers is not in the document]
mwbenowitz: We definitely don't expect last names/first names to be well labeled, the script is doing its best to identify them, and seems to be handling it okay.

I think in general we want this to be reusable, but it's hard because each citation can't be counted on to share any formatting. It would be great to have an element that can be passed a `style` prop like `<FormattedCitation style='apa' title={} {..etc}>` but I can't quite see how to achieve that with React, any thoughts?

I do think that it should only be passed the formatted logic.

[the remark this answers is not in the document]
mwbenowitz: That's with the caveat that <FormattedCitation> shouldn't just implement different methods for each style, but invoke the desired class and return that.

src/app/components/Citations/APACitation.jsx
mwbenowitz: I see what you're saying, I definitely don't disagree and will see what I can do

src/app/components/Citations/APACitation.jsx
mwbenowitz: yeah, that's definitely better, was just trying to distinguish from the input and what the user would see.

src/app/components/List/EditionsList.jsx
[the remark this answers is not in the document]
hokei: OH in terms of the `APA`, `MLA`, etc components being reuseable with each other, I think we should just ... not do that.  They might share methods to get a well formatted title or something, but I think it's perfectly OK (maybe better?) to have a separate `APACitation` and `MLACitation`, etc, etc.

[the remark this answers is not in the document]
hokei: The reuseability comment is more about how much this should be able to be ripped out of RN and still work, vs tied into the way RN works.

src/app/components/Citations/APACitation.jsx
```diff
@@ -0,0 +1,150 @@
...
+    this.edition = this.props.edition ? ` (${this.props.edition})` : '';
+    this.volume = this.props.volume ? ` (${this.props.volume})` : '';
+
+    this.link = this.props.sourceLink ? <a href={`https://${this.props.sourceLink}`}>{this.props.sourceLink}</a> : '';
+
+    this.govReportStatus = this.props.isGovernmentDoc;
+  }
+
+
+  formatAgentNames(agentType, nameFirst = false) {
```
hokei, line 36: A tip:  It'll be a lot easier to test functions like these if you move them out of the component (so you won't have to simulate the component and passing props) 
(so pass `this.props.agents` into the function)

src/app/components/Citations/APACitation.jsx
```diff
@@ -0,0 +1,150 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+class APACitation extends React.Component {
+  static getInitials(name) {
+    return name.trim().split(/\s+/).map(i => `${i.substr(0, 1).toUpperCase()}.`).join(' ');
+  }
+
+  componentWillMount() {
```
hokei, line 10: does `componentDidMount` not receive props?

src/app/components/SearchResults/ResultsList.jsx
```diff
@@ -83,17 +86,36 @@ class ResultsList extends React.Component {
...
+          <DS.SearchResultItem
+            id={`search-result-${result.uuid}`}
+            resultIndex={index}
+            headingContent={titleElement}
+            subtitleContent={EditionCard.getSubtitle(result.sub_title)}
+            authorLinkElement={authorLinkElement ? joinArrayOfElements(authorLinkElement, ', ') : undefined}
+            editionInfo={EditionCard.getEditionData(result, previewEdition, eReaderUrl, referrer, showRequestButton)}
+            editionsLinkElement={allEditionsLink}
+          />
+          {citationElement}
```
hokei, line 117: Instead of this, you can go `{shouldShowCitations && <APACitation {... all the props } />`

hokei:
The approach looks good.  You might want to break out some of these functions even more (or take them out of the component) if it turns out that testing them is difficult.  I'll leave it to you!

src/app/components/Citations/APACitation.jsx
mwbenowitz: I tried a few different ways of organizing this logic, and it feels like it needs to be here since this is so closely tied to displaying the APA format. I'll keep that in mind though if I can't get it to test cleanly.

src/app/components/Citations/APACitation.jsx
mwbenowitz: For some reason with `componentDidMount` everything was rendered as `undefined` and then the parsing function was run. But it didn't update the displayed component, maybe because I'm not assigning anything to `state` in the function?

appConfig.js
```diff
@@ -39,4 +39,7 @@ export default {
   feedback: {
     formURL: 'https://api.airtable.com/v0/appFLZEc3LmVZCRxn/Feedback',
   },
+  displayCitations: {
```
hokei, line 42: update the "FeatureFlag" portion of the readme?

src/app/components/Citations/APACitation.jsx
```diff
@@ -0,0 +1,186 @@
...
+        {this.link}
+      </>
+    );
+  }
+
+  // Returns appropriate APA citation version
+  render() {
+    const volumeData = this.volume !== '';
+    return (
+      <div className="apa-citation body-small">
```
hokei, line 145: can you add a little margin-bottom to this so that it's not stuck to the title of the next book?

hokei:
should citations not show at all on the edition details page (shouldn't they show in the data table?)

test/unit/CitationFormat.test.js
```diff
@@ -0,0 +1,180 @@
+/* eslint-disable no-unused-expressions */
+/* eslint-disable no-underscore-dangle */
+/* eslint-disable react/jsx-filename-extension */
+/* eslint-disable prefer-destructuring */
+/* eslint-env mocha */
```
hokei, line 5: I don't think you need anything above L5

test/unit/APACitation.test.js
```diff
@@ -0,0 +1,234 @@
+/* eslint-disable react/jsx-filename-extension */
```
hokei, line 1: you don't need this

test/unit/ResultsList.test.js
```diff
@@ -65,9 +66,11 @@ describe('Results List', () => {
     });
 
     describe('Complete Results Data', () => {
+      let resultsBlock;
       let resultsData;
       before(() => {
-        resultsData = component.instance().formatAllResultsData(results.data.works, 'eReaderUrl', 'Referrer')[0].props;
+        resultsBlock = component.instance().formatAllResultsData(results.data.works, 'eReaderUrl', 'Referrer')[0];
+        resultsData = resultsBlock.props.children[0].props;
```
hokei, line 73: why did this change?  seems ... disconcerting.

test/unit/CitationFormat.test.js
```diff
@@ -0,0 +1,180 @@
...
+
+    it('should set government document status', () => {
+      const citationData = CitationFormatter.getCitationData(work, work.editions[0]);
+      expect(citationData.isGovernmentDoc).to.equal(false);
+    });
+  });
+
+  describe('getAgentsOfType()', () => {
+    const testAgents = [
+      { name: 'test1', roles: ['test', 'other_again'] },
```
hokei, line 109: can we have these values be some kind of representation of what the data actually might be, to help it serve better as documentation?

src/app/components/Citations/APACitation.jsx
mwbenowitz: What's the best way to do this? I meant to ask because all the CSS lives in design system, right?

test/unit/CitationFormat.test.js
```diff
@@ -0,0 +1,180 @@
...
+      const testLink = CitationFormatter.setLinkFields(null);
+      expect(testLink.link).to.equal(null);
+      expect(testLink.link_date).to.equal(null);
+    });
+  });
+
+  describe('isGovernmentReport()', () => {
+    it('should return true if government_report measurement is 1', () => {
+      const testMeasures = [
+        { quantity: 'testing', value: 0 },
```
hokei, line 159: can you change this value to 1 so that we make sure the "government_document" value trumps all other values?  

(Alternately, can be a different test, but I don't think it's necessary)

test/unit/ResultsList.test.js
mwbenowitz: To put the citation with the card I wrapped them in an empty element, so this just follows that and gets the card as a child. Nothing changed beyond that so I think it's okay?

test/unit/APACitation.test.js
```diff
@@ -0,0 +1,234 @@
...
+          publishers: ['Test Publisher'],
+        }}
+        publicationYear="2020"
+        edition={null}
+        volume="Vol. 2"
+        sourceLink="test_link"
+        isGovernmentDoc={false}
+      />);
+      const citationText = component.find('.apa-citation').text();
+      // eslint-disable-next-line max-len
```
hokei, line 181: Why does this need to be disabled?  (If it's too long you can concatenate strings across 2 lines)

test/unit/APACitation.test.js
```diff
@@ -0,0 +1,234 @@
...
+          agents={{
+            authors: ['Author, Test'],
+            editors: ['Editor, Test'],
+            translators: ['Translator, Test'],
+            illustrators: ['Illustrator, Test'],
+            publishers: ['Test Publisher'],
+          }}
+          publicationYear="2020"
+          edition="Test Edition"
+          volume={null}
```
hokei, line 200: Can you leave out volume here?  It should do the same thing.

src/app/components/Citations/APACitation.jsx
hokei: Yeah I have thoughts on why this is, but I don't think doing it this way is going to cause problems and refactoring it so that `componentDidMount` will work is going to be another chunk of work so I think it's ok to leave it the way it is.

mwbenowitz:
> should citations not show at all on the edition details page (shouldn't they show in the data table?)

They should, but it means updating the API response to add more work-level metadata to the edition detail response, and sense we paused that discussion regarding author names I think it's best to resolve that before implementing this? Then we can create a ticket for updating the edition detail page with the data from however we decide to address that.

test/unit/APACitation.test.js
mwbenowitz: It doesn't I just added it while I was working, copied the code and then forgot to take it out.

test/unit/ResultsList.test.js
hokei: Why is it affecting components that don't have citations in them?

mwbenowitz: I thought it wasn't? I updated the `return` from `formatAllResultsData` that includes the citation if the flag is enabled, should I change that so if the flag isn't present it renders differently? This felt like the cleanest solution to me.

hokei: hmmm maybe I'm reading this wrong.  I thought this test was for a result without citation, so resultsData should return the same format as before, but here it's nesting a level down, even without the citation?  I don't think it's a huge deal, just stood out to me.

mwbenowitz: it does nest even without the citation. I put the `key` in a wrapping `div` so that when the citation is present React doesn't complain about not having a unique key for repeating elements. Maybe there's a more elegant way to handle this but I've stared at for a while now and can't figure out what that would be

src/client/styles/components/Citations.scss
```diff
@@ -0,0 +1,4 @@
+div.apa-citation {
+  margin: 0 0 1em 0;
+  font-size: 90%;
+}
```
hokei, line 4: nit

test/unit/ResultsList.test.js
hokei: aaah I'm so sorry I see what's happening now.  This is fine.

PR #174, opened 2020-05-19
HOTFIX AWS Elastic Beanstalk Environment Fix

mwbenowitz:
A new/udpated npm package is causing a conflict with the AWS EBS environment. It is outputting unexpected non UTF-8 characters during the install process and this causes deployments to fail. This fix, which is supposedly the official recommendation from AWS is to pipe the output of the NPM install process to /dev/null. This removes the ability to easily review installation logs for troubleshooting, but does allow for deployments. Review should be made in the future to see if this can be removed.

NO JIRA TICKET

- Patches AWS ElasticBeanstalk deployment process

-  No testing required
- Review config files for sanity

### DIFF
--- a/.ebextensions/00_environment.config
+++ b/.ebextensions/00_environment.config
@@ -2,3 +2,7 @@ option_settings:
   aws:elasticbeanstalk:application:environment:
     NODE_ENV: production
     APP_ENV: production
+
+container_commands:
+  replace_ebnode_containerfile:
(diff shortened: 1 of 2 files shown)


mwbenowitz:
haha yeah, it's definitely a pretty ugly hack, I'm not happy about it. I'll have to ask around to see just how much work it would really be to move. I do think in the long term it gives us the flexibility and control that we need. Especially since other issues have prevented us from bumping up the node version on these machines for a long time.

PR #176, opened 2020-05-29
SFR-935 when searching by author VIAF, prepopulate searchbar with aut…

hokei:
…hor name

[Jira Ticket](http://jira.nypl.org/browse/SFR-935)

- When clicking author name, author name prepopulates in searchbar.

-

### DIFF
--- a/test/unit/SearchHeader.test.js
+++ b/test/unit/SearchHeader.test.js
@@ -91,4 +91,57 @@ describe('SearchHeader', () => {
       expect(component.state('searchQuery').queries[0].query).to.equal('johnson');
     });
   });
+
+  describe('search term prepopulation', () => {
+    let component;
+    it('should prepopulate based on passed in value', () => {
+      component = mount(<SearchHeader initialQuery={{ queries: [{ field: 'author', query: 'melville' }] }} />);
+      expect(component.find('#searchBarId-input-textfield').prop('value')).to.equal('melville');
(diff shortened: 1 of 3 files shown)


test/unit/SearchHeader.test.js
```diff
@@ -91,4 +91,57 @@ describe('SearchHeader', () => {
...
+        initialQuery={{
+          queries: [{ field: 'viaf', query: '12345' }],
+          showQueries: [{ field: 'author', query: 'shakespeare' }],
+        }}
+      />);
+      expect(component.find('#searchBarId-input-textfield').prop('value')).to.equal('shakespeare');
+      expect(component.find('#dropdownId').prop('value')).to.equal('author');
+    });
+
+    it('should not prepopulate if viaf is not for author', () => {
```
mwbenowitz, line 125: I'm not sure this should be true. I think any time you click an agent name (be it author, publisher or something else like translator) and we execute a search it should the name you clicked. Right now I think that this doesn't matter since we are only enabling author searches in this way, but I'm not sure this test should be included.

hokei: The searchbar only has fields "keyword, author, subject, title" so there's no way to display "Publisher: Houghton" or whatever, so it will show up that you searched the publisher in the display heading, but in the search bar itself, it'll be empty.  
I think keeping the test here serves as clear documentation, and we want to change it we can change the test.

PR #181, opened 2020-06-02
No ref template fixes

hokei:
NO-JIRA

- Template work in Design System made it not necessary to use `DS.Container`.  This PR removes DS.Container and uses template styling.

-  the spacing errors in dev are gone: 
Screen Shot 2020-06-01 at 3 11 47 PM


hokei:
uhh a note on reviewing this cuz github isn't that great at diffing when tabbing on things get messed up:  `DS.Container` is no longer used, and all the pages don't have super wonky layouts (especially on Safari)

PR #185, opened 2020-06-12
SFR-937 Upgrade DS to 0.4.1 

hokei:

- upgrades SFR to use Design System 0.4.0
- New overrides to beautify

-  Check for regressions in front end app

### DIFF
--- a/src/app/components/About/About.jsx
+++ b/src/app/components/About/About.jsx
@@ -5,7 +5,7 @@ import * as DS from '@nypl/design-system-react-components';
 import { Link } from 'react-router';
 import Breadcrumbs from '../Breadcrumbs/Breadcrumbs';
 
-class License extends React.Component {
+class About extends React.Component {
   componentDidMount() {
     global.window.scrollTo(0, 0);
   }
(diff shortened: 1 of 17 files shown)


src/app/components/About/About.jsx
```diff
@@ -71,21 +71,21 @@ All the materials in Digital Research Books Beta are completely free to read and
                     The material in Digital Research Books Beta are drawn from several
                                         public sources, mainly
               {' '}
-              <DS.UnderlineLink url="https://www.hathitrust.org/">HathiTrust</DS.UnderlineLink>
+              <DS.Link href="https://www.hathitrust.org/">HathiTrust</DS.Link>
```
mwbenowitz, line 74: Is there a difference between the `href` and `url` attributes used on `DS.Link`. I see both and assume they do the same thing but maybe should be aligned for consistency?

hokei: url is never used now, it's always href.

mwbenowitz: Look down in the `License` file looks like its showing up there

hokei: omg the license page wasn't working at all, thank you for catching this.

PR #188, opened 2020-06-16
RENO-1463 Account for absolute links in footer

misselbeck:
[Jira Ticket](http://jira.nypl.org/browse/RENO-1463)

### The Problem:

The previous version of dgx-react-footer used relative urls for the NYPL footer links. For apps such as this one, which live on a different domain, those links don't work correctly.

The new version of dgx-react-footer adds a new prop called urlType which can optionally be passed into the Footer component to maintain absolute urls for the footer links.

This approach is similar to how the dgx-header-component handles links.

- Updates dgx-react-footer to v0.5.7
- Adds the new urlType prop to the Footer component in `Application.jsx`

-  The NYPL links in the footer section should now all be absolute urls and open as expected.

### Note:
I did not bump the version for the app, as I assumed this update will be packaged with other work that is in progress.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGE LOG
-##[0.9.2]
+## [0.9.3] 
+- Updated dgx-global-footer to 0.5.7.
+- Adds the new urlType prop to the Footer component set to "absolute"
+
+## [0.9.2]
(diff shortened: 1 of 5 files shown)


misselbeck:
Thanks, Crystal. I went ahead and bumped the app version, as requested.

PR #204, opened 2021-04-26
NOREF Make API Configurable

mwbenowitz:
The next.js application currently reads the API URL from a static configuration file committed to the repo. This makes temporary development/qa deployments difficult because the temporary API will generally be deployed to a non-standard URL (frequently a bare IP address).

This update updates the application to use a variable `API_URL` to configure this. It can be set in the `.env` file for local development or provided as a `-e API_URL=url` flag when running the the application in a container. This increases the portability of the containerized application and for a broader array of deployments.

This also eliminates the use of the hard coded application environment variable in the `SearchApi.ts` file. This configuration can be simplified as a) the API endpoints should be static regardless of deployment and b) the API URL itself can be configured via the environment.

To run the app in a container with this new configuration the command would be `docker run -e APP_ENV='env' -e API_URL='url' -p 3000:3000 sfr-front-end`

-  Validate that the application still runs properly when run locally and in a container. (Both run by me, but could use verification)
- Prior to merging this PR the AWS CloudFormation stack will have to be updated to use these environment variables


.env.sample
```diff
@@ -1,4 +1,8 @@
-NYPL_API_BASE_URL
-NYPL_OAUTH_URL
-NYPL_OAUTH_KEY
-NYPL_OAUTH_SECRET
\ No newline at end of file
+# GENERAL CONFIGURATION
+APP_ENV=development
+
+# API ENDPOINT
+API_URL=localhost:5000
```
hokei, line 5: let's put a non-local API url here for easy copy-pasting in the future

hokei: Could we make this an optional variable?  So if one isn't passed, it's derived from `APP_ENV` (localhost for dev, the qa api for qa, etc)

mwbenowitz: that's a good idea I think (for both comments). Easy enough to do and I think that QA and dev would point to the same URL

hokei (approved):

PR #206, opened 2021-04-27
No ref cleanup

hokei:

- Better error pages
- Loading and error has correct margins

-


hokei:
@mwbenowitz It looks like the api doesnt' quite work when you do a search, here?

mwbenowitz:
@hokei This is strange the `API_URL` should be set to `localhost` but it's querying `localhost:3000` instead. I think there should be a simple fix to the issue I just need to figure out what that might be. (The API itself works, it's just the connection between the two applications)

PR #208, opened 2021-04-28
HOTFIX Specify API port

mwbenowitz:
The front-end application is calling the local API instance, but on the incorrect port, this attempts to fix the issue by statically specifying that port.

Verify that the app is able to query the API


hokei (commented):
PS if you add the `no-changelog` label it won't error on changelog

mwbenowitz:
thanks for the tip! yeah I think it failed because I'm screwing around with the services right now to try and update it in a weird way. Should never happen in "real" PRs

mwbenowitz:
@hokei Now it's ready to review! The environment in the latest comment should work for querying, etc.

PR #212, opened 2021-04-29
HOTFIX Update branch tags

mwbenowitz:
This commit includes one fix and one improvement to the QA deployment process.

The fix is to correct a regression that was causing the `latest` tag for the application container to be deployed. The updated `task-definition` was not being passed in the correct step

The improvement is to tag images with the branch name instead of the git hash, this makes them more identifiable.


mwbenowitz:
This should resolve the issue with changes not propagating to the QA environments. In the new task definition there is an image tag that has the branch name in it. This is what is build from the latest code and deployed: https://console.aws.amazon.com/ecs/home?region=us-east-1#/taskDefinitions/sfr-pipeline-testing/30

PR #230, opened 2021-06-02
NOREF Add production build step to GHA

mwbenowitz:
- Adds a production build action that is triggered against merges to the production branch (automated production deployments)

-  This can't be tested and should be verified to work properly before the formal launch on 6/9 (aka merge this along with a functional change into production and be sure that the functional change deploys properly)

### DIFF
--- a/.github/workflows/build-production.yaml
+++ b/.github/workflows/build-production.yaml
@@ -1,0 +1,50 @@
+# On merge to production,
+# build a container and deploy to ECR
+name: Publish Production
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


hokei (approved):

hokei:
Not sure what the lint error is considering you're not updating the lockfile, but try running npm i again and see if it solves it

mwbenowitz:
I ran it again and it passed, I think it was some transient issue with GHA itself because I saw a few other actions fail similarly around the same time. Regardless, will merge this in!

PR #231, opened 2021-06-09
Uses absolute urls in header

hokei:

- Fixes links in header so that they route to `nypl.org`
- Looks like the current production branch doesn't have a bunch of the header changes from recently (it's using the cdn instead of the react component) so I've copied those changes over from `development`

-


mwbenowitz (approved):
👍

PR #250, opened 2021-06-29
NOREF Update ES endpoint

mwbenowitz:
As part of handling production traffic loads on DRB the elasticsearch client changed, and the new endpoint needs to be provided to the test environments here.

-  Validate that the test environment can query the API


hokei (approved):
LGTM

PR #264, opened 2021-07-22
Re-adds missing class to search results

hokei:

- Adds sidebar class to main by adding a prop into the `<Layout />` component
- This eliminates the need to define a `<main>` component in each page. 

-


mwbenowitz (approved):
This looks good to me and works in the test environment.

The only question I have is if we want to peg the design system module to accept only patch updates or let it stick with the minor updates. I know this wasn't the issue here but might give us more confidence in the stability of things?

PR #273, opened 2021-09-13
Add EDD button

Toxiapo:
[Jira Ticket](http://jira.nypl.org/browse/SFR-944)

- Adds EDD button to the edition page if the item can be found at NYPL
- If the patron is logged in, display a Request button that takes you to the Research Catalog EDD request form, display a Login button if not.
- Adds nodejs and npm version to package.json and npmrc

-  To test the cookie functionalities, pull this branch and [alias](https://setapp.com/how-to/edit-mac-hosts-file) your localhost to *.nypl.org (e.g local.nypl.org), the login does not save the cookie in your browser unless it's coming from *.nypl.org


src/components/EditionCard/EditionCard.tsx
```diff
@@ -1,16 +1,21 @@
...
+import { MediaTypes } from "~/src/constants/mediaTypes";
+
+export const NYPL_SESSION_ID = "nyplIdentityPatron";
 
 export const EditionCard: React.FC<{ edition: WorkEdition; title: string }> = ({
   edition,
   title,
 }) => {
+  const [cookies] = useCookies([NYPL_SESSION_ID]);
+
```
Toxiapo, line 17: The example from react-cookie wraps the whole app with CookiesProvider, but it seems to work without it in our case, maybe because that's only used for setting cookies?

CHANGELOG.md
```diff
@@ -1,31 +1,50 @@
 # CHANGE LOG
 
+## [0.10.5]
```
hokei, line 3: let's minor bump this to 0.11.0 as EDD is a new feature.

src/components/EditionCard/EditionCard.test.tsx
```diff
@@ -88,3 +91,40 @@ describe("Edition Year with Minimal Data", () => {
     expect(screen.queryByText("Read Online")).not.toBeInTheDocument();
   });
 });
+
+describe("Edition with EDD", () => {
```
hokei, line 95: Add additional tests for Edition with both EDD and regular read-online/download (should show read-online/download)

src/components/EditionCard/EditionCard.tsx
hokei: yes, I think that's correct.

src/util/EditionCardUtils.tsx
```diff
@@ -267,4 +268,41 @@ export default class EditionCardUtils {
...
+
+  static gedEddLink(eddLink: ItemLink, isLoggedIn: boolean) {
+    if (isLoggedIn) {
+      return (
+        <>
+          You can request a partial scan via NYPL&nbsp;
+          <a
+            href="https://www.nypl.org/research/scan-and-deliver"
+            target="_blank"
+            rel="noreferrer"
```
hokei, line 280: why noreferrer?

Toxiapo: ahh, it seems like my VS linter automatically added that because I have `target="_blank"`... I can use `<Link>` instead so it doesn't complain.

hokei:
Looks like for this book:  82c82a19-749b-4a59-a4e1-d4c6d0c32b94 The featured edition is the Scan and Deliver edition, rather than a read online edition.  Can we figure out if this is because of something in the front end or the back end?

src/components/EditionCard/EditionCard.test.tsx
hokei: These test names are a bit difficult to read and parse.  
Can we say, for each test, what the data is that you're checking, and what you expect?  
"Shows Download and Read Online button when edition has both EDD and readable links", 
"Shows Login button when EDD is available but user is not logged in", etc etc etc

src/util/EditionCardUtils.tsx
hokei: yeah I missed that - you should use `Link` instead of `a` for links always, if you're using NYPL design system

Toxiapo:
> Looks like for this book: 82c82a19-749b-4a59-a4e1-d4c6d0c32b94 The featured edition is the Scan and Deliver edition, rather than a read online edition. Can we figure out if this is because of something in the front end or the back end?

I don't think we ever checked the media type, we only checked the availability of the item and return the first match. (EDD or Readable). I can push a commit here to update this, should be straightforward I hope.

mwbenowitz:
On the edd v read online `media_type` question I think Jiayong is right, but this means that we need to make sure we pass the right edition in the search results page.

This is also only an issue when we have multiple editions from the same year, one with EDD and one with a read online option. I could set the system to sort by "preferred" delivery method (RO => EDD => None) and then by publication_date which would make it so that the preferred edition is always Read Online if available

Toxiapo:
> On the edd v read online `media_type` question I think Jiayong is right, but this means that we need to make sure we pass the right edition in the search results page.
> 
> This is also only an issue when we have multiple editions from the same year, one with EDD and one with a read online option. I could set the system to sort by "preferred" delivery method (RO => EDD => None) and then by publication_date which would make it so that the preferred edition is always Read Online if available
 
If the data can be sorted in the system that would be great!

src/components/EditionCard/EditionCard.test.tsx
Toxiapo: Updated, thanks!

hokei (approved):
This looks really good!  Looks like that issue with that one book is a backend one, so @mwbenowitz can you link that ticket when you make it?  

Otherwise, let's get this into QA!

mwbenowitz:
@hokei @Toxiapo yup, as discussed we will add a new way of identifying catalog records, which will enable us to fix the issue identified above. This is incidental to this work and should not block merging into QA: [SFR-1317](http://jira.nypl.org/browse/SFR-1317)

PR #274, opened 2021-09-22
Add EDD button to the edition detail page

Toxiapo:
[Jira Ticket](http://jira.nypl.org/browse/SFR-944)

- Added the missing EDD button to the edition detail page.
- Use the `item.link.flags` tag to determine the media type.  

-

### DIFF
--- a/src/components/EditionDetail/Edition.tsx
+++ b/src/components/EditionDetail/Edition.tsx
@@ -19,7 +19,6 @@ const Edition: React.FC<{ editionResult: EditionResult }> = (props) => {
   const { pathname, query } = router;
   const edition: ApiEdition = props.editionResult.data;
 
-  //FIXME: Same logic applies here as 'Work.tsx' page
   const featuredInstance = edition.instances.find(
     (edition: any) =>
       edition.items &&
--- a/src/components/EditionDetail/Edition.tsx
+++ b/src/components/EditionDetail/Edition.tsx
@@ -19,7 +19,14 @@ const Edition: React.FC<{ editionResult: EditionResult }> = (props) => {
   const { pathname, query } = router;
   const edition: ApiEdition = props.editionResult.data;
 
-  const featuredInstance = edition.instances[0];
+  //FIXME: Same logic applies here as 'Work.tsx' page
+  const featuredInstance = edition.instances.find(
+    (edition: any) =>
+      edition.items &&
+      edition.items.length &&
+      edition.items[0].links &&
+      edition.items[0].links.length
+  );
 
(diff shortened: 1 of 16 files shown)


src/util/EditionCardUtils.tsx
```diff
@@ -183,6 +183,7 @@ export default class EditionCardUtils {
   static getReadOnlineLink = (item: ApiItem) => {
     const getReadLink = (item: ApiItem, mediaType: string) => {
       if (!item || !item.links) return undefined;
+      /// TODO: feature `flags` object indicator
```
hokei, line 186: nit: Perhaps call this "use `item.link.flags`" so it's not easily confused with the "Feature Flags" feature, which has been on our to-do list for a while.

src/util/EditionCardUtils.tsx
```diff
@@ -304,4 +343,13 @@ export default class EditionCardUtils {
       );
     }
   }
+
+  // Get any link except for catalog
+  static getPreviewItem(items: ApiItem[] | undefined) {
```
hokei, line 348: In the Search Results page, does this method get called, or does it first apply the filter?  eg:  When you filter for "PDFs only", the preview items should be the ones with PDFs.

Toxiapo: This seems fine because it performs the search filter before calling the method.

src/components/EditionDetail/Edition.tsx
```diff
@@ -19,7 +19,14 @@ const Edition: React.FC<{ editionResult: EditionResult }> = (props) => {
   const { pathname, query } = router;
   const edition: ApiEdition = props.editionResult.data;
 
-  const featuredInstance = edition.instances[0];
+  //FIXME: Same logic applies here as 'Work.tsx' page
```
hokei, line 22: elaborate on this?

src/components/Work/Work.tsx
```diff
@@ -19,6 +19,8 @@ const WorkDetail: React.FC<{ workResult: WorkResult }> = (props) => {
   //Edition Card Preprocessing
   const authorsList = EditionCardUtils.getAuthorsList(work.authors);
 
+  // FIXME: This will show 'EDD' until Mike updates the backend to change the sorting (RO => EDD => None)
```
hokei, line 22: This isn't a front-end issue, so there's no need to mention it here.  To the extent that this isn't the desired functionality when hooked up to a real back-end, that's documented in tickets and in confluence.

hokei (approved):
Looks good!

PR #275, opened 2021-09-22
SFR-1235 Integrating Webreader

hokei:

- Swaps out the current webpub-viewer for the new webreader.  
- Updates unit tests to use V2 of the reader API
- Adds an optional env variable to set PDF proxy.  

-


src/pages/read/[linkId].tsx
```diff
@@ -2,14 +2,17 @@ import React from "react";
 import ReaderLayout from "~/src/components/ReaderLayout/ReaderLayout";
 import { readFetcher } from "~/src/lib/api/SearchApi";
 import { LinkResult } from "~/src/types/LinkQuery";
+
 export async function getServerSideProps(context: any) {
+  console.log("linkId", context.query);
```
Comment (hokei) [inline src/pages/read/[linkId].tsx:7]: somehow `context.query.linkId` is getting the link (http://localhost:3000/read/6551603) as well as pdf.min.js.  This is throwing an error `Error: cannot find work with linkId pdf.worker.min.js` 

Something needs to be set up such that it looks in the right place?

src/components/ReaderLayout/ReaderLayout.tsx
```diff
@@ -3,21 +3,26 @@ import * as DS from "@nypl/design-system-react-components";
...
+  const origin = router.basePath;
   const link: ApiLink = props.linkResult.data;
-
+  const url = formatUrl(link.url);
   const edition = link.work.editions[0];
 
-  const isRead = link.flags.reader;
-  const isEmbed = link.flags.embed;
+  const isEmbed = MediaTypes.embed.includes(link.media_type);
+  const isRead = MediaTypes.read.includes(link.media_type);
```
Toxiapo, line 25: Should we also update the `getReadOnlineLink` function so that it matches this? 

(http://34.204.45.18:3000/work/24211e77-3713-4ca9-9c0c-0c0c6cc54cc8 this work doesn't render because of the inconsistent media type checking)

src/constants/mediaTypes.ts
```diff
@@ -1,7 +1,7 @@
 //Read should have `application/pdf` as well but the reader doesn't support it yet
```
Toxiapo, line 1: nit: Outdated comment. We now support PDFs!

src/components/ReaderLayout/ReaderLayout.tsx
hokei: Oh this is a bug!  This should match getReadOnlineLink, I think!

Toxiapo:
This looks great to me,  I am excited to see this on QA! 🎉

PR #298, opened 2021-10-29
Upgrading to node 16

hokei:

- Upgrades DRB to use Node 16
- Removes V1 webreader 
- Removes all dependencies on Enzyme.  
- Uses `next-router-mock` instead of our custom mock script, which has better features.

-


src/components/Feedback/Feedback.test.tsx
```diff
@@ -56,21 +56,23 @@ describe("Feedback", () => {
...
+    window.fetch = fakeFetch;
     test("should invoke method when success and feedback set", () => {
       jest.spyOn(window, "alert").mockImplementation(() => jest.fn());
 
       const feedback = screen.getByRole("textbox", {
         name: "Comments (Required)",
       });
       const yesRadio = screen.getByLabelText("Yes");
-
+      expect(true).toBeTruthy;
```
samanthaandrews, line 68: stray assertion i think?

samanthaandrews, line 68: stray assertion i think?

src/__tests__/Search.test.tsx
```diff
@@ -19,6 +15,10 @@ import { FilterLanguagesCommonTests } from "./componentHelpers/FilterLanguages";
 import { FilterFormatTests } from "./componentHelpers/FilterFormats";
 import { findFiltersForField } from "../util/SearchQueryUtils";
 import filterFields from "../constants/filters";
+import mockRouter from "next-router-mock";
+
+jest.mock("next/router", () => require("next-router-mock"));
```
samanthaandrews, line 20: is this how we are mocking the next/router? just curious; wasn't sure what this line was doing but seems to be working

samanthaandrews (changes requested):
Just one small comment and then we are good to approve! I pulled in the branch locally to test that I could npm install, build the app, and run tests. Everything is working like a charm 💫

src/__tests__/Search.test.tsx
hokei: Yes - we used to use `MockNextRouter.tsx` but it was hacked together, so using an actual library seemed more maintainable

samanthaandrews (approved):

PR #300, opened 2021-11-02
Upgrades web-reader to 0.2.5

samanthaandrews:
No Jira Ticket

- Updates @nypl/web-reader package so that PDF rendering performance enhancements in scrolling mode can be tested by QA

-  Load a large single-resource PDF in Firefox, Chrome, or Safari. The resource should render in scrolling mode without issue. Browser should not crash.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "sfr-bookfinder-front-end",
-  "version": "0.11.1",
+  "version": "0.11.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


hokei:
What is the improvement that should be visible by QA?  I've tried loading in the test environment and it seems like it's still pretty slow?  Do we have numbers to show that it's easier, like comparing with prod?  (Is it useful to compare with prod when prod has more resources than QA?)

samanthaandrews:
> What is the improvement that should be visible by QA? I've tried loading in the test environment and it seems like it's still pretty slow? Do we have numbers to show that it's easier, like comparing with prod? (Is it useful to compare with prod when prod has more resources than QA?)

Rendering a large PDF won’t crash a user’s browser with this change. Loading a large PDF has not been changed in any way. It still takes a long time to load, but quickly renders one page as it becomes visible in the viewport.

hokei:
Ah ok!  Can you give an example of a PDF that crashed the browser before, but doesn't now?  eg: http://3.82.127.202:3000/read/4552285 -- I started loading it last night and this morning it's still on "pdf is loading", with nothing in the console.  Might need a new ticket on messaging.

samanthaandrews:
I am getting cors errors in console when loading those PDFs in the preview environment
Screen Shot 2021-11-03 at 8 50 00 AM

Have you seen this before? My changes wouldn't have changed how we are fetching PDFs... But it does seem necessary to fix before QA tests anything new. I'd like for QA to test the original resources that were mentioned in the ticket.

hokei:
x-posting from slack: Oh yes!  Sorry I totally forgot that the test deploys currently aren't being accepted by the proxy and that we're waiting till the tugboat migration to fix it.

(aka I should have known this sorry)

hokei (approved):
For QA, we should still try to find a pdf that fails in drb-prod and see if that same pdf fails.  (They'll have different link IDs and QA has fewer books than prod)

samanthaandrews:
I'm not sure if this is the production deployment or not, but I did find the same resource there that we are testing in QA? https://drb-production.nypl.org/read/4549091

PR #305, opened 2022-01-13
Fix extra scroll bars

Toxiapo:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1371)

- Upgrade webreader to 0.3.3
- CSS update to fix vertical scroll bar 

-

### DIFF
--- a/styles/main.scss
+++ b/styles/main.scss
@@ -56,7 +56,6 @@ body {
 }
 
 .nypl-ds {
-  overflow-x: hidden;
   fieldset {
     border: none;
   }


styles/main.scss
```diff
@@ -41,6 +41,12 @@ $fontFamily: "system-ui";
 
 @import "./dividers.scss";
 
+*,
+*::before,
+*::after {
+  box-sizing: border-box;
+}
+
```
Toxiapo, line 49: It looks like I still need this line : ( The vertical scrollbar only shows up on a big monitor.

styles/main.scss
```diff
@@ -50,7 +56,6 @@ body {
 }
 
 .nypl-ds {
-  overflow: unset;
```
Toxiapo, line 53: Do we need this? I think we do need the overflow to be hidden(default) or it shows a horizontal bar on the landing page.

hokei:
@mwbenowitz can we figure out why this didn't trigger a test deploy?  Is that functionality that got reverted somehow and we missed it?

Toxiapo:
This is the preview link from the dashboard, I think we just need to fix the auto-share feature. (I don't have the admin to see the settings, it takes me to a broken page after I click on the settings button)

https://pr305-aic8bhujmbqyp9sxs67v9lhj3oqevhne.tugboat.qa/

styles/main.scss
hokei: I think it's ok for now.  This definitely will be a thing we're changing soon with the DS upgrade

hokei (approved):
This looks great!  I checked the Tugboat settings and fiddled with some things - we probably can't check it until a new PR gets built.  So, let's merge this in and see if the tugboat's fixed when we prepare this for release?

PR #327, opened 2022-05-02
[WIP] SFR-1464: Update search unit tests

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1464)

-

-

### DIFF
--- a/src/__tests__/Search.test.tsx
+++ b/src/__tests__/Search.test.tsx
@@ -50,14 +50,14 @@ describe("Renders Search Results Page", () => {
     });
   });
 
-  test("Digital Research Books Beta links to homepage", () => {
-    const homepagelinks = screen.getAllByRole("link", {
-      name: "Digital Research Books Beta",
-    });
-    homepagelinks.forEach((link) => {
-      expect(link).toHaveAttribute("href", "/");
-    });
-  });
+  // test("Digital Research Books Beta links to homepage", () => {
+  //   const homepagelinks = screen.getAllByRole("link", {
+  //     name: "Digital Research Books Beta",
+  //   });
+  //   homepagelinks.forEach((link) => {
+  //     expect(link).toHaveAttribute("href", "/");
+  //   });
+  // });
   test("DRB Header is shown", () => {
     expect(
       screen.getByRole("heading", { name: "Digital Research Books Beta" })
@@ -70,7 +70,7 @@ describe("Renders Search Results Page", () => {
   });
   test("Main Content shows the current search query with 'alert' role", () => {
     expect(screen.getByRole("alert")).toHaveTextContent(
-      "Search results for keyword: Animal Crossing"
+      'Search results for keyword: "Animal Crossing"'
     );
   });
   test("Item Count shows correctly", () => {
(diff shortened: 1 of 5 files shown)


src/__tests__/Search.test.tsx
```diff
@@ -86,21 +86,28 @@ describe("Renders Search Results Page", () => {
     test("clicking 'filters' button shows filters contents", () => {
       fireEvent.click(screen.getByText("Filters (0)"));
       expect(
-        screen.getByRole("combobox", { name: "Items Per Page" })
+        screen.getAllByRole("combobox", { name: "Items Per Page" })[0]
```
hokei, line 89: why is there more than one combobox with "items per page"?

hokei (changes requested):
I'm concerned about the change from `getByRole` to `getAllByRole` - it implies that the modals are mounted even though they're not visible.  Is it good practice to have stuff that's rendered but invisible?  The idea is that a screenreader user shouldn't need to know the difference between `items-per-page` and `items-per-page-modal` - there isn't a meaningful difference between the two.  My instinct is that the mobile filters should have the same ids as the desktop filters, and stuff should only be mounted when they're visible - but my instincts might be outdated.  Thoughts?

src/__tests__/Search.test.tsx
jackiequach: Due to how Chakra loads responsive styling, the elements were accessible through the tests. I changed how the display property is populated by setting a state for the breakpoint width (this is how it is done on design system).

src/__tests__/componentHelpers/FilterFormats.tsx
```diff
@@ -1,7 +1,7 @@
 import { screen, within } from "@testing-library/react";
 
 export const FilterFormatTests = () => {
-  const formats = screen.getByRole("group", { name: "Format" });
+  const formats = screen.getAllByRole("group", { name: "Format" })[0];
```
hokei, line 4: Is this the same `getAllByRole` thing?

src/__tests__/componentHelpers/FilterLanguages.tsx
```diff
@@ -17,9 +17,7 @@ export const FilterLanguagesCommonTests = (
   });
 
   test("Language Filters shows all available languages", () => {
-    fireEvent.click(
-      screen.getAllByRole("button", { name: "Filter Languages" })[0]
-    );
+    fireEvent.click(screen.getByRole("button", { name: "Filter Languages" }));
```
hokei, line 20: How did this just work?  Wasn't there two components that were mounted?  How could both the previous test and this test pass with no other changes?

hokei: If the "unmount the unseen component" task is best separated out into a different ticket, then let's keep the `getAllByRole` tests and file that ticket.

jackiequach: The changes were made in Search.tsx to use states instead of Chakra's built in responsive styling since there is currently no way to test Chakra's styling.

hokei (approved):

PR #331, opened 2022-06-03
SFR-1424: Update format filters

jackiequach:

- Change format filters to "Readable", "Downloadable", and "Requestable"

-


hokei (approved):
oh wow so the way this works is, it's just passed to the query parameter and then passed directly to the backend, and there's no changes in the search logic needed?  That's kind of amazing!

jackiequach:
> probably a different ticket unless its suuuper eas Screen Shot 2022-06-06 at 12 19 13 PM 

@hokei I believe Sean created a separate ticket for this. The issue stems from the DS upgrade.

PR #332, opened 2022-06-06
Sfr 1475 reservoir

hokei:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1475)

- Upgrades SFR to Reservoir

-


src/__tests__/Search.test.tsx
```diff
@@ -127,19 +127,20 @@ describe("Renders Search Results Page", () => {
     describe("Per Page filters", () => {
       test("Changes Sort By sends new search", () => {
         const allSorts = screen.getAllByLabelText("Items Per Page");
-        const wideSorts = allSorts[0];
-        expect(wideSorts).not.toBeVisible();
-        const modalSorts = allSorts[1];
+        const modalSorts = allSorts[0];
```
hokei, line 130: looks the order of these changed.  The proper way is to get the section with the modal, and then go `withinModal(getByLabelText("items per page"))` but that would require adding a heading to the modal.  As it is right now I'm not sure if this would pass A11Y, but FilterBar is going to be out soon so it might not matter.

jackiequach: For now, I think this is fine for testing the modal functionality

jackiequach (approved):
Changes look good!

PR #337, opened 2022-07-08
SFR-1446: Implement feature flags

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1446)

- Adds ability to define and create feature flags in code
- Turn them on/off through query parameters starting with "featureFlag_"
- Query parameters are persistent for the browser session

-

### DIFF
--- a/src/context/FeatureFlagContext.tsx
+++ b/src/context/FeatureFlagContext.tsx
@@ -1,0 +1,59 @@
+import { useRouter } from "next/router";
+import React, { createContext, useContext, useEffect, useState } from "react";
+import { extractQueryParam } from "../util/LinkUtils";
+import { flattenDeep } from "../util/Util";
+
+type FeatureFlagContextType = {
+  activeFeatureFlags: string[];
+  setActiveFeatureFlags: (activeFeatureFlags: string[]) => void;
+  isFlagActive: (flag: string) => boolean;
+};
+
+export const FeatureFlagContext =
+  createContext<FeatureFlagContextType>(undefined);
+
+export const FeatureFlagProvider: React.FC = ({ children }) => {
+  const [activeFeatureFlags, setActiveFeatureFlags] = useState<string[]>([]);
+  const isFlagActive = (flag: string) => {
+    return activeFeatureFlags.includes(flag);
+  };
+
+  const router = useRouter();
+  const { query } = router;
+
+  useEffect(() => {
+    const storedFeatureFlagsStr = sessionStorage.getItem("features");
+    let storedFeatureFlags: string[] = [];
+    if (storedFeatureFlagsStr) {
+      try {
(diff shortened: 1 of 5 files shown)


src/context/FeatureFlagContext.tsx
```diff
@@ -0,0 +1,86 @@
+import { useRouter } from "next/router";
+import { ParsedUrlQuery } from "querystring";
+import React, { createContext, useContext, useEffect, useState } from "react";
+import { extractQueryParam } from "../util/LinkUtils";
+
+interface FeatureFlag {
```
hokei, line 6: Any reason why `interface` is preferred here over `type` ?

jackiequach: Originally used this to define props for `FeatureFlagProvider`, will change this to `type`. According to the [TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces), they are mostly interchangeable and based on personal preference.

hokei: Great --  I'm happy to continue using `type` by default so that uses of `interface` signal that we specifically need those features, aka when we're expecting people to extend the interfaces.

src/context/FeatureFlagContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const [featureFlags, setFeatureFlags] = useState<FeatureFlag>({});
+  const isFlagActive = (flag: string) => {
+    return featureFlags[flag];
+  };
+
+  const router = useRouter();
+  const { query } = router;
+
+  useEffect(() => {
+    const storedFeatureFlagsStr = sessionStorage.getItem("featureFlags");
```
hokei, line 40: Can you explain how the functionality interacts with the session storage?  It seems from the tests that it's all based on what the query parameter is, so why is it looking into session storage at all?

src/util/__tests__/LinkUtils.test.tsx
```diff
@@ -31,3 +37,25 @@ describe("Generate back to serach url", () => {
...
+
+describe("Extracts query parameter from url", () => {
+  test("extractQueryParam returns a single flag", () => {
+    mockRouter.setCurrentUrl(
+      "https://drb-qa.nypl.org/edition/1780467?feature=new_feature"
+    );
+    const features = extractQueryParam(mockRouter.query, "feature");
+    expect(features).toEqual("new_feature");
+  });
+  test("extractQueryParam returns undefined if multiple flags", () => {
```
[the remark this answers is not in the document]
jackiequach: This test is named incorrectly. The case would be if there were multiple flags of the same name with different values, i.e. feature_totalCount=true&feature_totalCount=false

hokei (approved):

PR #341, opened 2022-08-02
SFR-1505: Fix back to search results bug

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1505)

- Changes the check for "search" to the endpoint "/search" to prevent links without the endpoint from being recognized as a search results link for prod, i.e. digital-re**search**-books-beta.nypl.org

-


Toxiapo (approved):
🎉

PR #342, opened 2022-08-03
Fix QA Deploy

samanthaandrews:
My last merge to `development` branch failed the QA deploy action. This PR fixes the issue and ensures it doesn't happen in the future.

- Updates package-lock file to ensure dependencies are installed correctly on CI
- Run the Docker build step in the CI action so that it runs on each PR. This won't catch ECS issues, but it will at least ensure that the docker image is successfully built on each PR so that a failure like this won't happen in the future.


Dockerfile
```diff
@@ -19,7 +19,7 @@ ENV NEXT_PUBLIC_AIRTABLE_API_KEY $airtable_api_key
 ENV NEXT_PUBLIC_READER_VERSION=v2
 
 # Install dependencies.
-COPY package.json ./
+COPY package.json package-lock.json ./
```
jackiequach, line 22: Is it necessary to copy package-lock.json? It will be updated automatically after npm install is run.

samanthaandrews: It will be updated automatically, but that is where we will likely run into issues. Without copying it here, it will do a completely fresh install of all packages, which means in most cases, since we are using the `^version` caret pattern in package.json, it will update to all future minor/patch versions, without incrementing the major version. This can cause issues, and it was for me in my testing (specifically I was seeing a lot of type errors in the `docker build` step that I wasn't getting when running locally). For this reason it is best practice to include the package-lock file in all build steps.

jackiequach: Got it, thanks for the clarification

jackiequach (approved):

PR #351, opened 2022-09-20
SFR-1535: Add clear filters button

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1535)

- Adds Clear Filters button to search results page

-


src/components/Search/Search.tsx
```diff
@@ -274,6 +274,28 @@ const SearchResults: React.FC<{
...
+                width: "100%",
+                display: {
+                  base: "block",
+                  md: "none",
+                },
+              }}
+            >
+              Clear Filters
+            </Button>
+          ) : null}
```
hokei, line 298: a shorthand: you can replace this with the logical `&&` operator https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

src/components/Search/Search.tsx
```diff
@@ -274,6 +274,28 @@ const SearchResults: React.FC<{
...
           />
+          {searchQuery.filters.length > 0 ? (
+            <Button
+              id="clear-filters-button"
+              buttonType="secondary"
+              type="reset"
+              onClick={() => {
+                changeFilters([]);
+              }}
+              __css={{
```
hokei, line 285: If you need a version of the "secondary" button that is blue, perhaps this is worth asking if the Design System should have one or have a way of creating a blue "secondary" button.

jackiequach: spoke to Marty, this variant is in the works for Reservoir and will be included in the next release on Thursday.

hokei (approved):
just a small stylistic thing, but then we're good!

PR #354, opened 2022-10-03
SFR-1533: Include selected languages in filters

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1533)

- Add selected languages to the available options when there are no results for the selected languages

-


hokei (changes requested):
As this is an edge case that I imagine we're not going to check every time, we should have a test for this.  

Also - and this is a style/implementation thing - as we've recently added a new filter type, have you considered an implementation that first loads all of the selected filters, and then adds in all of the available options from data (instead of this implementation, which first looks at the filter-able options in the data and then adds the selected stuff if they're not already in there).  

I'm thinking about this because, in this implementation, if we wanted to add any new filter times we would have to first remember that this bug is possible with that new filter, and remember to specifically handle it if it is.  Whereas, if we handle all the applied filters up front, we don't need to worry about if it's a "disappearing" filter or not.

jackiequach:
I think it would be good to implement a Playwright test for this particular edge case since it would be an E2E test utilizing both frontend and backend data. 

I see what you mean for loading the selected filters first, but since we get count for each language from the backend, the selected filter would need to be replaced with the available option to populate the count. We would still need to check if it's a "disappearing" filter.

@hokei Please let me know your thoughts on this and if I am misunderstanding any of it.

hokei (changes requested):
That makes sense, re: the progressive filters.  

For the test, recapping what we talked about in our 1:1 -- try adding a mock for the fetch request and making it return the minimal fixture that would have caused this erroneous behavior.  Then you can assert that this behavior doesn't happen.  In general we don't want to use E2E tests for edge cases because they're still pretty expensive to execute.

hokei (approved):

PR #356, opened 2022-10-11
adding playwright tests for item detail page

chrismulholland:

-

-

### DIFF
--- a/playwright/tests/itemDetailPageElements.spec.ts
+++ b/playwright/tests/itemDetailPageElements.spec.ts
@@ -1,0 +1,127 @@
+import { Given, Then } from "@cucumber/cucumber";
+import { expect } from "@playwright/test";
+import { CustomWorld } from "../support/setup";
+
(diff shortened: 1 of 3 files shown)


playwright/tests/itemDetailPageElements.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+Then("I see the authors heading", async function (this: CustomWorld) {
+  await expect(
+    this.page.locator("//dt[contains(text(),'Authors')]")
+  ).toBeVisible();
+});
+Then("I see the author", async function (this: CustomWorld) {
+  await expect(
+    this.page.locator(
+      "//dt[contains(text(),'Authors')]/following-sibling::dd/a[contains(text(),'Library of Congress')]"
```
samanthaandrews, line 70: This xpath seems like it could be brittle but I understand that there might not be a better solution.

jackiequach (approved):

PR #357, opened 2022-10-14
Edd playwright

chrismulholland:

-

-

### DIFF
--- a/playwright/tests/edd.spec.ts
+++ b/playwright/tests/edd.spec.ts
@@ -1,0 +1,34 @@
+import { Given, Then, When } from "@cucumber/cucumber";
+import { expect } from "@playwright/test";
+import { CustomWorld } from "../support/setup";
+// require("dotenv").config();
+// import dotenv from "dotenv";
+// import path from "path";
+
+Given("I have done a keyword search", async function (this: CustomWorld) {
+  await this.page.goto(`${this.parameters.appUrl}`);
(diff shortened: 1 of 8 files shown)


playwright/tests/edd.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { Given, Then, When } from "@cucumber/cucumber";
+import { expect } from "@playwright/test";
+import { CustomWorld } from "../support/setup";
+import dotenv from "dotenv";
+dotenv.config();
```
samanthaandrews, line 5: Ok this is almost what we need, but because the values you are adding are secrets (username and PIN), we should change things slightly.

You should change your env file to be named `.env.local` and add `.env.local` to the .gitignore file so that it is not committed to git -- I actually think this is already in this repository's gitignore file, but check to make sure.

Then you will change this line to be `dotenv.config({ path: "env.local" })`

A regular `.env` file is used for environment variables that are not secrets, and that file is usually committed to git.

samanthaandrews (changes requested):
Just one small change. Re-request review from me when you're finished and I'll approve 👍 Also, you'll need to move this PR out of its current "Draft" state.

playwright/tests/edd.spec.ts
chrismulholland: done.

samanthaandrews (approved):

PR #365, opened 2022-11-08
SFR-1557: Automate tag and release

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1557)

- Add job to build-production.yaml to automate tag and release

-


.github/workflows/build-production.yaml
```diff
@@ -7,6 +7,38 @@ on:
...
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          strategy: package
+          prefix: v
+          # use the body of the PR commit as the tag message
+          tag_message: ${{ github.event.pull_request.body }}
+
+      - name: Release
+        id: create_release
+        if: steps.autotagger.outputs.tagname != ''
+        uses: actions/create-release@v1.0.0
```
samanthaandrews, line 32: Note: This `actions/create-release` is currently archived and not maintained, I think it's ok to use it, but in the future we might consider moving over to one of the maintained actions mentioned in the docs https://github.com/actions/create-release

samanthaandrews (approved):

.github/workflows/build-production.yaml
jackiequach: created a Jira ticket here: https://jira.nypl.org/browse/SFR-1568

PR #374, opened 2022-11-21
Release 0.16.0 to production

jackiequach:
- Add github actions to automate tag and release
- Replace DS Header component with dgx-header-component v2.8.0-r16-rc-2


jackiequach:
Same regression tests failing as prior release

samanthaandrews (approved):

PR #378, opened 2022-12-22
Pub year filter playwright

chrismulholland:

-

-


playwright/features/pub_yearFilter.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Publication Year filter test
+
+    Scenario: As a user I should be able to filter searches by publication year and see relevent results
+        Given I want to filter by publication year
```
samanthaandrews, line 4: I'm thinking we should change these statements slightly so they make a little more sense to someone reading the tests for the first time, like `Given I am on the home page` and then `When I search for New York` and then `Then I filter by publication years 1900-1900`

playwright/tests/pub_yearFilter.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  async function (this: CustomWorld) {
+    return await this.page.goto(`${this.parameters.appUrl}`);
+  }
+);
+
+When(
+  "I search for New York to test publication year filter",
+  async function (this: CustomWorld) {
+    await this.page.locator('[aria-label="Item Search"]').fill("New York");
+    await this.page.locator("#searchbar-button-search-bar").click();
```
samanthaandrews, line 16: Is there a locator here we could use that isn't an id? Like a button within a form that includes the word "Search"?

playwright/tests/pub_yearFilter.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+);
+
+Then("I input publication years 1900-1900", async function (this: CustomWorld) {
+  await this.page
+    .locator('[aria-describedby="date-filter-from-helperText"]')
```
samanthaandrews, line 22: Let's use the name attribute since it's more human readable. `[name="Date From"]`

playwright/tests/pub_yearFilter.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    .fill("1900");
+  await this.page
+    .locator('[aria-describedby="date-filter-to-helperText"]')
```
samanthaandrews, line 25: Same here, `[name="Date To"]`

samanthaandrews (changes requested):
A couple of changes, otherwise this looks good. Go ahead and re-request review when you've made updates.

playwright/features/pub_yearFilter.feature
chrismulholland: good suggestion.  done.

playwright/tests/pub_yearFilter.spec.ts
chrismulholland: done.

playwright/tests/pub_yearFilter.spec.ts
chrismulholland: done.

playwright/tests/pub_yearFilter.spec.ts
samanthaandrews: should be `this.page.getByRole(...)`

we have to use `this.<playwright_api_element>` because we are wrapping playwright in Cucumber which is happening in the `setup.ts` file.

chrismulholland: sorry, typo from me.  I actually am using 'this'.
image
await this.page.getByRole("button", { name: "Search" }).click();

chrismulholland: interesting how vscode spots the error even before I run it.

chrismulholland: I'm wondering if there is an additional library import that needed since 'getByRole' is foreign to vscode.

samanthaandrews: This is happening because we need to upgrade Playwright to version 1.29.1. I attempted this but ran into other issues when trying to upgrade. For now, we will leave this locator as is and I'll create another ticket to work on updating Playwright.

samanthaandrews (approved):

PR #381, opened 2023-01-13
SFR-1587: Implement collection list

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1587)

- Adds type definitions for OPDS2 Feed
- Replaces Search Examples with Recently Added Collections on the landing page
- Adds tests for `CollectionList` and `CollectionCard`

-

### DIFF
--- a/src/lib/api/CollectionApi.ts
+++ b/src/lib/api/CollectionApi.ts
@@ -16,8 +16,12 @@ export const collectionFetcher = async (query: CollectionQuery) => {
   const res = await fetch(url.toString());
 
   if (res.ok) {
-    const collectionResult: CollectionResult = await res.json();
-    return collectionResult;
+    try {
+      const collectionResult: CollectionResult = await res.json();
+      return collectionResult;
+    } catch (e) {
+      throw new Error(e.error);
+    }
   } else {
     throw new Error(
       `cannot find collection with identifier ${query.identifier}`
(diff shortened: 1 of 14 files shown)


src/lib/api/CollectionApi.ts
```diff
@@ -0,0 +1,26 @@
...
+  const collectionApiQuery = {
+    ...query,
+  };
+
+  const url = new URL(collectionUrl + "/" + query.identifier);
+  url.search = new URLSearchParams(collectionApiQuery.toString()).toString();
+  const res = await fetch(url.toString());
+
+  if (res.ok) {
+    const collectionResult: CollectionResult = await res.json();
```
samanthaandrews, line 19: nit: it could be a good idea to wrap this `await res.json()` in a try/catch block just in case something goes awry with the json parsing

src/util/CollectionUtils.tsx
```diff
@@ -0,0 +1,27 @@
+import { formatUrl } from "./Util";
+import { PLACEHOLDER_COVER_LINK } from "../constants/editioncard";
+import { MediaTypes } from "../constants/mediaTypes";
+import { Opds2Feed, OpdsLink } from "../types/OpdsModel";
+
+export default class CollectionUtils {
```
samanthaandrews, line 6: nice!

samanthaandrews (approved):
Looks great!! Not sure of the process for visual QA on DRB, but you could send the Vercel preview URL to Apoorva if you want to. Probably not necessary since y'all are using Reservoir components, but it's an option.

PR #384, opened 2023-01-24
SFR-1586: Add primary hero to landing page

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1561)

- Add primary hero to landing page
- Removes `<main>` from `Layout` since it is included in DS `Template` used on every page

-

### DIFF
--- a/src/components/Landing/Landing.tsx
+++ b/src/components/Landing/Landing.tsx
@@ -7,11 +7,12 @@ import {
   Hero,
   Link,
   Template,
+  TemplateAboveHeader,
   TemplateBreakout,
   TemplateContent,
   TemplateContentPrimary,
-  TemplateContentTop,
   TemplateFooter,
+  TemplateHeader,
 } from "@nypl/design-system-react-components";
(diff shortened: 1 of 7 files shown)


src/components/Landing/Landing.tsx
```diff
@@ -22,41 +23,65 @@ const LandingPage: React.FC<{ collections?: Opds2Feed }> = ({
...
+        world-wide--all free to read, download, and keep. No library card
+        required. This is an early beta test, so we want your feedback!{" "}
+        <Link href="/about" color="ui.link.primary">
+          Read more about the project
+        </Link>
+        .
+      </span>
+      <Box marginTop="s">
+        Search the World's Research Collections
+        <SearchForm isFullWidth />
```
samanthaandrews, line 38: is there a way we can add the appropriate style without needing to pass down a prop? maybe by wrapping the `<SearchForm /> ` in a `<Box />` and adding additional style properties there

samanthaandrews (approved):

PR #388, opened 2023-03-07
SFR-1588:  Add Collection page

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1588)

- Add [collectionId] page endpoint and corresponding Collection component

-

### DIFF
--- a/src/components/Collection/Collection.test.tsx
+++ b/src/components/Collection/Collection.test.tsx
@@ -6,6 +6,7 @@ import { CollectionQuery, CollectionResult } from "~/src/types/CollectionQuery";
 import {
   collectionData,
   collectionWithPagination,
+  emptyCollectionResult,
 } from "~/src/__tests__/fixtures/CollectionFixture";
 import mockRouter from "next-router-mock";
 import userEvent from "@testing-library/user-event";
@@ -171,5 +172,25 @@ describe("Render Collection Page with >10 items", () => {
       expect(previousLink).not.toBeInTheDocument();
       expect(nextLink).not.toBeInTheDocument();
     });
+  });
+});
+
+describe("Renders No Results when no results are shown", () => {
+  beforeEach(() => {
+    render(
+      <Collection
+        collectionQuery={collectionQuery}
+        collectionResult={emptyCollectionResult}
+      />
+    );
+  });
+
+  test("Shows the current collection with title", () => {
+    expect(
+      screen.getByText("Collection - Baseball: A Collection by Mike Benowitz")
+    ).toBeInTheDocument();
+  });
+  test("Item Count shows correctly", () => {
+    expect(screen.getByText("Viewing 0 items")).toBeInTheDocument();
   });
 });
--- a/src/components/Collection/Collection.test.tsx
+++ b/src/components/Collection/Collection.test.tsx
@@ -1,11 +1,14 @@
 import React from "react";
-import { act, fireEvent, screen } from "@testing-library/react";
+import { fireEvent, screen } from "@testing-library/react";
 import { render } from "../../__tests__/testUtils/render";
-import { resizeWindow } from "../../__tests__/testUtils/screen";
 import Collection from "./Collection";
 import { CollectionQuery, CollectionResult } from "~/src/types/CollectionQuery";
-import { collectionData } from "~/src/__tests__/fixtures/CollectionFixture";
+import {
+  collectionData,
+  collectionWithPagination,
+} from "~/src/__tests__/fixtures/CollectionFixture";
 import mockRouter from "next-router-mock";
+import userEvent from "@testing-library/user-event";
 
 jest.mock("next/router", () => require("next-router-mock"));
 
@@ -13,7 +16,7 @@ const collectionResults: CollectionResult = collectionData;
 const collectionQuery: CollectionQuery = {
   identifier: "id",
   page: 1,
-  perPage: 10,
+  perPage: 5,
   sort: "relevance",
 };
 
@@ -25,9 +28,6 @@ describe("Renders Collection Page", () => {
         collectionResult={collectionResults}
       />
     );
-    act(() => {
-      resizeWindow(300, 1000);
-    });
   });
   test("Shows the current collection with title", () => {
     expect(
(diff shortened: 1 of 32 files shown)


src/util/CollectionUtils.tsx
```diff
@@ -24,4 +34,208 @@ export default class CollectionUtils {
...
+            Request
+          </Link>
+        </>
+      );
+    } else {
+      return (
+        <>
+          May be available via NYPL<br></br>
+          <Link
+            to={`https://login.nypl.org/auth/login?redirect_uri=${encodeURIComponent(
```
samanthaandrews, line 230: Nit: consider moving this login url out into a constant

src/util/CollectionUtils.tsx
```diff
@@ -24,4 +34,208 @@ export default class CollectionUtils {
...
+
+  static getEddLinkElement(eddLink: OpdsLink, isLoggedIn: boolean) {
+    if (isLoggedIn) {
+      return (
+        <>
+          <Box whiteSpace="initial">
+            You can request a partial scan via NYPL
+          </Box>
+          <Link
+            to="https://www.nypl.org/research/scan-and-deliver"
```
samanthaandrews, line 210: Nit: consider moving this login url out into a constant

src/util/CollectionUtils.tsx
```diff
@@ -24,4 +34,208 @@ export default class CollectionUtils {
...
+        </Link>
+      );
+    }
+
+    return undefined;
+  };
+
+  static getNoLinkElement(showRequestButton: boolean) {
+    if (showRequestButton) {
+      return <span>Not Yet Available {showRequestButton}</span>;
```
samanthaandrews, line 134: Does this render a boolean value of true or false?

src/util/CollectionUtils.tsx
```diff
@@ -24,4 +34,208 @@ export default class CollectionUtils {
...
+    }
+    return <>Languages: Undetermined</>;
+  }
+
+  // Rights
+  static getLicense(rights: Rights) {
+    return rights ? `License: ${rights.rightsStatement}` : "License: Unknown";
+  }
+
+  static getReadLink = (links: OpdsLink[], type: "readable" | "embedable") => {
```
samanthaandrews, line 103: Since it seems like the `type: "readable" | "embedable"` is specific to what we can read/open in this app, we might want to make is a named type that we can reuse. Something like `export type ReadOnlineTypes = "readable" | "embedable"`

src/components/Collection/Collection.tsx
```diff
@@ -0,0 +1,245 @@
...
+import CollectionUtils from "~/src/util/CollectionUtils";
+import ResultsSorts from "../ResultsSorts/ResultsSorts";
+import { useRouter } from "next/router";
+import { CollectionItemCard } from "../CollectionItemCard/CollectionItemCard";
+import Link from "../Link/Link";
+
+const Collection: React.FC<{
+  collectionQuery: CollectionQuery;
+  collectionResult: CollectionResult;
+}> = (props) => {
```
samanthaandrews, line 42: Nit: we can destructure props here, i.e. `({ collectionResult, collectionQuery })` and then further destructure `collectionResult` within the component.

samanthaandrews (changes requested):
I left a couple comments! Overall I need to see two things:

1. Convert functions that return JSX that are called directly inside a component into React functional components and render them directly.
2. Unit tests for sorting and pagination. Ideally we could also do this with integration tests using Cypress or Playwright, but for now, let's create unit tests. Specifically let's make sure we cover potential edge cases.

src/components/Collection/Collection.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+describe("Renders Collection Page", () => {
+  beforeEach(() => {
+    render(
+      <Collection
+        collectionQuery={collectionQuery}
+        collectionResult={collectionResults}
+      />
+    );
+    act(() => {
+      resizeWindow(300, 1000);
```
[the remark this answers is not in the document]
jackiequach: Copied this from another test file, it was being used for mobile view. Removed in #389

src/util/CollectionUtils.tsx
jackiequach: Copied from `EditionCardUtils`, but looks like `showRequestButton` is never true. Removed for this.

src/util/CollectionUtils.tsx
```diff
@@ -24,4 +34,208 @@ export default class CollectionUtils {
...
+  };
+
+  static getNoLinkElement(showRequestButton: boolean) {
+    if (showRequestButton) {
+      return <span>Not Yet Available {showRequestButton}</span>;
+    }
+    return <>Not yet available</>;
+  }
+  static selectDownloadLink = (links: OpdsLink[]) => {
+    if (links) return undefined;
```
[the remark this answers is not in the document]
jackiequach: Yes, good catch.

samanthaandrews (changes requested):
Hey @jackiequach! I started reviewing this again but realized that I would be able to articulate what I'm looking for by coding it up in a separate branch, which now has an open PR here https://github.com/NYPL/sfr-bookfinder-front-end/pull/390/files

That PR has a lot of failing checks, which makes sense because I refactored it quickly without actually running the code 😅 I'm currently unable to run this project locally at all (I think due to the fact that I'm on a mac with an M1 chip).

I'm thinking it makes sense for you to pull down that branch and see if you can run it, get the tests to pass, and explore the refactoring that I did. The big takeaway is that I converted anything that renders JSX into a React component, and kept everything in CollectionUtils to only include functions that return text, OpdsLinks, etc. that would eventually get rendered inside a component. I think this makes the code easier to reason about and easier to test.

I also pulled out the various components that were part of CollectionItemCard into their own files, as that component was getting large and it needed to be more readable. This is more of a stylistic preference on my end.

src/components/CollectionList/CollectionList.tsx
```diff
@@ -13,9 +13,15 @@ export const CollectionList: React.FC<{ collections: Opds2Feed }> = ({
           <SimpleGrid gap="l" columns={4}>
             {collections.groups.map((collection, index) => {
               if (index < 8)
```
samanthaandrews, line 15: Let's turn this magic number 8 into a named constant so we know why it is specifically set to 8

src/components/Collection/Collection.tsx
```diff
@@ -0,0 +1,245 @@
...
+          backgroundColor="section.research.primary"
+          heroType="tertiary"
+          heading={
+            <Heading level="one" id="tertiary-hero">
+              <>
+                Digital Research Books <sup>Beta</sup>
+              </>
+            </Heading>
+          }
+        />
```
samanthaandrews, line 162: Seems like this is a component that is duplicated in a couple of other places in the code. Let's pull it out into a separate component and re-use it in all the places it belongs.

src/components/Collection/Collection.tsx
```diff
@@ -0,0 +1,245 @@
...
+                  {pub.metadata.subtitle && (
+                    <Box marginBottom="xs">{pub.metadata.subtitle}</Box>
+                  )}
+                  {pub.metadata.creator && (
+                    <Box marginBottom="xs">
+                      By <Author author={pub.metadata.creator} />{" "}
+                    </Box>
+                  )}
+                  <CollectionItemCard collectionItem={pub}></CollectionItemCard>
+                </Box>
```
samanthaandrews, line 226: Since this component is quite large and complex, let's pull this out into a separate component and put it in a new file.

samanthaandrews (approved):

PR #389, opened 2023-03-10
SFR-1588: Revise tests from collection page branch

jackiequach:

- Address test comments from #388 
- Clean up collection fixture for reusability

-


samanthaandrews (approved):
LGTM! One optional optimization would be to also mock/spy on the API calls that happen when paginating or sorting, and confirm that those calls are being made with the correct parameters.

PR #390, opened 2023-03-14
Sam's CollectionItemCard edits: Refactors collection item card and collection utils

samanthaandrews:
- Refactors CollectionUtils to include only functions that return OpdsLinks, text, or anything that isn't actually JSX
- Turns anything that renders JSX into a react component
- Separates all of the CollectionItemCard components into separate files for readability and easier testing

-  I didn't actually run this code while I was refactoring, so @jackiequach if you wouldn't mind running this branch to make sure everything runs as expected and looks ok!

### DIFF
--- a/src/components/CollectionItemCard/LanguageDisplayText.tsx
+++ b/src/components/CollectionItemCard/LanguageDisplayText.tsx
@@ -1,0 +1,14 @@
+import React from "react";
+import { Box } from "@nypl/design-system-react-components";
+
+const LanguageDisplayText: React.FC<{
+  language: string;
+}> = ({ language }) => {
+  return (
+    <Box as="p">
+      {language ? `Languages: ${language}` : "Languages: Undetermined"}
+    </Box>
+  );
+};
+
+export default LanguageDisplayText;
--- a/src/components/CollectionItemCard/PublisherAndLocation.tsx
+++ b/src/components/CollectionItemCard/PublisherAndLocation.tsx
@@ -1,0 +1,21 @@
+import React from "react";
+import { Box } from "@nypl/design-system-react-components";
+import CollectionUtils from "~/src/util/CollectionUtils";
+
+const PublisherAndLocation: React.FC<{
+  pubPlace: string;
+  publisher: string;
+}> = ({ pubPlace, publisher }) => {
+  const displayLocation = CollectionUtils.getPublisherDisplayLocation(pubPlace);
+  const displayName = CollectionUtils.getPublisherDisplayText(publisher);
+
+  return (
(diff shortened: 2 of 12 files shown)


src/components/Collection/Collection.tsx
```diff
@@ -49,25 +50,22 @@ const Collection: React.FC<{
...
-  const collectionId = CollectionUtils.getId(collectionResult.links);
-
-  const collectionMetadata = collectionResult.metadata;
-  const itemsPerPage = collectionMetadata.itemsPerPage;
-  const totalItems = collectionMetadata.numberOfItems;
-  const lastPageLink =
-    collectionResult.links[collectionResult.links.length - 1].href;
+  const { metadata, links, publications } = collectionResult;
+  const { itemsPerPage, numberOfItems, currentPage, title, description } =
+    metadata;
```
samanthaandrews, line 55: object destructuring here to help us write less `metadata.title`, `metatada.description`, etc. later in the component

src/components/CollectionItemCard/LanguageDisplayText.tsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { Box } from "@nypl/design-system-react-components";
+
+const LanguageDisplayText: React.FC<{
+  language: string;
+}> = ({ language }) => {
+  return (
+    <Box as="p">
```
samanthaandrews, line 8: Wasn't sure about the accessibility rules for putting text directly in a div, so I added this `as` prop, but it might change the display. I'll leave it up to you @jackiequach if this should be a p tag or just a div.

src/components/CollectionItemCard/PublisherAndLocation.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const PublisherAndLocation: React.FC<{
+  pubPlace: string;
+  publisher: string;
+}> = ({ pubPlace, publisher }) => {
+  const displayLocation = CollectionUtils.getPublisherDisplayLocation(pubPlace);
+  const displayName = CollectionUtils.getPublisherDisplayText(publisher);
+
+  return (
+    <Box as="p">
```
samanthaandrews, line 13: @jackiequach same here

src/components/CollectionItemCard/LanguageDisplayText.tsx
jackiequach: I think this is better for accessibility, but you will need to add `margin="0"` to fix the spacing change.

jackiequach (approved):
Looks good, just minor comments for display

PR #398, opened 2023-05-02
SFR-1675: Playwright tests on homepage collections

clarissarichard:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1675)

- Adds Playwright tests to check the collections section on the homepage

-  Running Playwright tests and all tests pass

### DIFF
--- a/playwright/tests/homePageElements.spec.ts
+++ b/playwright/tests/homePageElements.spec.ts
@@ -73,14 +73,46 @@ Then("I see advanced search link", async function (this: CustomWorld) {
   await expect(this.page.locator("text=Advanced Search")).toBeVisible();
 });
 
-// 1. write assertion to find Recently Added Collections heading
+// 1. Write assertion to find Recently Added Collections heading
 // 2. Confirm heading is visible
 Then("I see the collections heading", async function (this: CustomWorld) {
-  await expect(this.page.getByRole('heading', {name: 'Recently Added Collections'})).toBeVisible();
+  await expect(this.page.getByRole('heading', { name: 'Recently Added Collections' })).toBeVisible();
 })
 
 // 1. Find a locator that is not dependent on user content
 // 2. Assert that locator is visible
+// Things we could assert: 
+// - there's at least one card
+// - the card has the text "Collection"
+// - the card is a link
+// - the card has the text "Items"
+// - the card has an image
+Then("I see the list of collection cards", async function (this: CustomWorld) {
+  const collectionCountLocator = await this.page.locator("//b[contains(text(), 'Collection')]").count(); // count number of collection cards
+  await expect(collectionCountLocator).toBeGreaterThan(0); // at least one card exists with the text "Collection"
+
+  // use firstCollection Card to condense code - not working bc of hierarchy
+  //const firstCollectionCard = await this.page.getByRole("heading", { name: "Recently Added Collections" }).first(); // find first card
+  //await expect(firstCollectionCard.filter({ has: this.page.getByRole("link") })).toHaveAttribute("href", /collection/); // first card is a link
+  //await expect(firstCollectionCard.getByRole("img")).toBeVisible(); // first card has an image
+  
+  // confirm first card is a link, has an image, and has "Items" text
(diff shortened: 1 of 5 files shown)


playwright/tests/homePageElements.spec.ts
```diff
@@ -73,6 +73,47 @@ Then("I see advanced search link", async function (this: CustomWorld) {
   await expect(this.page.locator("text=Advanced Search")).toBeVisible();
 });
 
+// 1. Write assertion to find Recently Added Collections heading
+// 2. Confirm heading is visible
```
samanthaandrews, line 77: Now that the test is written and passing, let's remove these todo list comments

playwright/tests/homePageElements.spec.ts
```diff
@@ -73,6 +73,47 @@ Then("I see advanced search link", async function (this: CustomWorld) {
...
+})
+
+// 1. Find a locator that is not dependent on user content
+// 2. Assert that locator is visible
+// Things we could assert: 
+// - there's at least one card
+// - the card has the text "Collection"
+// - the card is a link
+// - the card has the text "Items"
+// - the card has an image
```
samanthaandrews, line 89: Let's remove these comments as well

playwright/tests/homePageElements.spec.ts
```diff
@@ -73,6 +73,47 @@ Then("I see advanced search link", async function (this: CustomWorld) {
...
+  await expect(this.page.getByRole("img", { name: "Collection" }).first()).toBeVisible(); // first card has an image
+  await expect(this.page.getByRole("heading", { name: /Items/ }).first()).toBeVisible(); // first card has "Items" text
+  
+  // confirm correct link is located
+  const linkLocator = await this.page.getByRole("link", { name: "Collection" }).first().getAttribute("href");
+  console.log("collection card href attribute:", linkLocator);
+
+  // confirm correct image is located
+  const imgLocator = await this.page.getByRole("img", { name: "Collection" }).first().getAttribute("src");
+  console.log("collection card img src attribute:", imgLocator);
```
samanthaandrews, line 110: remove stray console.log

playwright/tests/homePageElements.spec.ts
```diff
@@ -73,6 +73,47 @@ Then("I see advanced search link", async function (this: CustomWorld) {
...
+
+  // confirm correct "Items" text is located
+  const itemsLocator = await this.page.getByRole("heading", { name: /Items/ }).first().innerText();
+  console.log("collection card items text:", itemsLocator);
```
samanthaandrews, line 114: remove stray console.log

samanthaandrews (changes requested):
Just a couple of small cleanup things 👍 This looks really good!

clarissarichard:
@samanthaandrews thanks for reviewing my PR! I removed the console.logs and todo comments, then pushed my changes. From the failing checks, I have a few questions to better understand what to correct:
- Even though we've already included a bullet point describing the new collections tests, do I also need to note that I removed the comments in the change log or should that check be ignored in this case?
- It looks like lint is recommending to change single quotes to double quotes. Which of the two are preferred here?
- I believe the Test & Coverage check found an error with an Advanced Search test. If this happens on a file that I did not make changes to, how should I proceed?

samanthaandrews:
> Even though we've already included a bullet point describing the new collections tests, do I also need to note that I removed the comments in the change log or should that check be ignored in this case?

Nope, you just need to update the Changelog once. It should include the changes that are part of the whole PR that will be merged to the main branch. I don't see any changes to the Changelog in this branch though. Not sure what happened to it, but you'll want to add a line there and push it up again in order to get that check to pass.

> It looks like lint is recommending to change single quotes to double quotes. Which of the two are preferred here?

Yeah I think our linter is configured to prefer double quotes. Within VS Code you can install an extension called `eslint` that will call out these errors for you locally so you can fix them before pushing up to this branch again. Lmk if you need help getting this to work!

> I believe the Test & Coverage check found an error with an Advanced Search test. If this happens on a file that I did not make changes to, how should I proceed?

This test failure is expected because it also exists on my base branch. Don't worry about it!

Great work.

PR #399, opened 2023-05-10
SFR-1682: Refactor page visits in Playwright tests

clarissarichard:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1682)

- Created a single assertion that accepts a page URL as a parameter

- Any page visit action in a Cucumber/Playwright test will use the same function
- Run Playwright tests and all tests pass

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,21 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Launch Program",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "program": "${workspaceFolder}\\playwright\\tests\\homePageElements.spec.ts",
+            "preLaunchTask": "tsc: build - tsconfig.json",
+            "outFiles": [
+                "${workspaceFolder}/dist/**/*.js"
+            ]
+        }
+    ]
+}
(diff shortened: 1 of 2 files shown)


clarissarichard:
After refactoring these tests, there are a couple things to note:
- `npm run cucumber -- playwright/features/edd.feature` is currently failing during the `Then I expect to see delivery options` test step with `Error: Timed out 5000ms waiting for expect(received).toBeVisible()`. I'm currently working on resolving the error.
- I did not refactor `helloWorld.feature` as it looks like a test file, but let me know if it's actually being run!

samanthaandrews:
I will do a full review of this tomorrow 👍 For now, addressing your comments:

> npm run cucumber -- playwright/features/edd.feature is currently failing during the Then I expect to see delivery options test step with Error: Timed out 5000ms waiting for expect(received).toBeVisible(). I'm currently working on resolving the error. Should be resolved now by overriding the default timeout!

Let me know if you need help!

> I did not refactor helloWorld.feature as it looks like a test file, but let me know if it's actually being run!

You can actually just delete this file and any associated specs. It was just a practice test I wrote when I was setting up Playwright and Cucumber in DRB.

.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
+{
```
samanthaandrews, line 1: Can you add this file to the `.gitignore` file? We will also need to make sure it gets removed from this PR. Can't remember off the top of my head how to do this! But maybe something like `git rm <filename>`? https://stackoverflow.com/questions/1139762/ignore-files-that-have-already-been-committed-to-a-git-repository

samanthaandrews (changes requested):
Just that one comment about the file we need to remove from git tracking. Otherwise this looks great!

samanthaandrews:
I added @ayan1229 to this PR review as well. @ayan1229 probably a good idea for you to look at these Playwright PRs that Clarissa is making. She's doing some of the refactoring work to make these tests look like the ones in OEB Web.

samanthaandrews (approved):
🚀

ayan1229 (approved):
Looks good! 
some changes can be made to make actions and assertions more dynamic. We can do it later

PR #401, opened 2023-05-17
SFR-1694: Refactor click actions in Playwright tests

clarissarichard:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1694)

- Created a single function that accepts a locator as a parameter and clicks on it
- Refactored Cucumber/Playwright tests so that all click actions/events use the same function

-  Run Playwright tests and all tests pass

### DIFF
--- a/playwright/support/mappings.ts
+++ b/playwright/support/mappings.ts
@@ -23,7 +23,8 @@ export const pages: { [name: string]: Pages } = {
 
 export const elements = {
   "advanced search link": "[href='/advanced-search']",
-  "search button": "#submit-button"
+  "advanced search button": "#submit-button",
+  "search button": "#searchbar-button-search-bar"
 };
 
 // trying to use built in locators
--- a/playwright/support/mappings.ts
+++ b/playwright/support/mappings.ts
@@ -33,7 +33,8 @@ export const elements = {
   "Latin language checkbox": "//label[contains(.,'Latin')]",
   "year filter button": "#year-filter-button",
   "first read online button": "text=Read Online >> nth=0",
-  "first request button": "text=Request >> nth=0"
+  "first request button": "text=Request >> nth=0",
+  "English language checkbox": "//*[@id='languages-checkbox-group-0-wrapper']/label/span[1]"
 };
 
 // trying to use built in locators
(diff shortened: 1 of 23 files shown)


playwright/support/mappings.ts
```diff
@@ -18,3 +18,22 @@ export const pages: { [name: string]: Pages } = {
...
+  "first login for options button": "text=Log in for options >> nth=0",
+  "government documents checkbox":
+    "//span[contains(text(), 'Show only US government documents')]",
+  "Latin language checkbox": "//label[contains(.,'Latin')]",
+  "year filter button": "#year-filter-button",
+  "first read online button": "text=Read Online >> nth=0",
+  "first request button":
+    "[href='https://www.nypl.org/research/collections/shared-collection-catalog/hold/request/b10715506-i13895605']",
+  "English language checkbox":
+    "//*[@id='languages-checkbox-group-0-wrapper']/label/span[1]",
```
samanthaandrews, line 37: Is there potentially a better/less potentially flaky locator we could access here? (this is arguably the hardest part of these Playwright tests, and i know in many instances, there just isn't something better!)

For this one in particular, I'm thinking something like `//input[text()='English']` might work. Lmk what you find.

samanthaandrews (approved):
I left one comment to see if we can try to get that locator replaced with something better, but overall this looks good and I think it's good to merge as is! Feel free to try to replace that locator with something else or not, up to you.

samanthaandrews:
This is a really great refactor. So much cleanup!!

playwright/support/mappings.ts
clarissarichard: Yes! I've been looking into optimizing these as well. I had changed some elements but not all of them. I'll also update the "Latin language checkbox" when I find something that works.

clarissarichard: @samanthaandrews I tried a few combinations (`input` looks to be more of a sibling to the text) and the two CSS locators that seem to work are `span:text('English')` and `[id='languages-checkbox-group-0-wrapper']`. Unless `span` and `id` are the flaky parts of the locator?

ayan1229 (approved):

PR #409, opened 2023-06-12
SFR-1703: Refactor element visibility assertions

clarissarichard:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1703)

- Created a single function that accepts an element as a parameter and expects it to be visible
- Refactored Cucumber/Playwright tests so that all visibility assertions use the same function

-  Run Playwright tests and all tests pass


samanthaandrews (approved):
My absolute favorite PRs are the ones where whole files get deleted! This is awesome 🚀

PR #410, opened 2023-06-13
SFR-1690: Implement adobe analytics

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1690)

- Replaces Google Analytics with Adobe Analytics implementation

-


.github/workflows/build-qa.yml
```diff
@@ -39,8 +39,12 @@ jobs:
           IMAGE_TAG: ${{ github.sha }}
           AIRTABLE_KEY: ${{ secrets.AIRTABLE_API_KEY }}
           NEW_RELIC_LICENSE_KEY: ${{ secrets.NEW_RELIC_LICENSE_KEY }}
+          NEXT_PUBLIC_ADOBE_ANALYTICS: ${{ secrets.NEXT_PUBLIC_ADOBE_ANALYTICS }}
```
samanthaandrews, line 42: this value doesn't need to be a secret, it's publicly available by looking at the DOM, so I would just inline it here

samanthaandrews, line 42: this value isn't a secret, so i would just inline it here

.github/workflows/build-production.yaml
```diff
@@ -71,8 +71,12 @@ jobs:
           IMAGE_TAG: ${{ github.sha }}
           AIRTABLE_KEY: ${{ secrets.AIRTABLE_API_KEY }}
           NEW_RELIC_LICENSE_KEY: ${{ secrets.NEW_RELIC_LICENSE_KEY }}
+          NEXT_PUBLIC_ADOBE_ANALYTICS: ${{ secrets.NEXT_PUBLIC_ADOBE_ANALYTICS }}
```
samanthaandrews, line 74: same here, but using the prod value

samanthaandrews (approved):

PR #411, opened 2023-06-15
SFR-1711: Remove header/footer from web reader page 

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1711)

- Hides the header/footer components if the web reader is displayed on /read endpoints.
- Removes type="module" from the Header/Footer scripts following updates from nypl-header-app.

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add TOC for single resource PDFs
 - Refactored Playwright/Cucumber tests to use new dropdown action function
 - Refactored Playwright/Cucumber tests to use new text fill function
+- Hide Header and Footer components on /read pages using the Web Reader
 
 ## [0.17.1]
 - Replace dgx-header-component and DS Footer with Header/Footer components from nypl-header-app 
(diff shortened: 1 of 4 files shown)


samanthaandrews (changes requested):
I think a cleaner, more sustainable way of doing this might be to re-compose the layout of the app so that the NYPL Header & Footer never wrap the Reader. I think this means we will need a parent <Layout /> component that wraps _all_ pages, including the Reader, and maybe a `<Page />` component that wraps all other pages.

```
<Layout>
  <Reader />
<Layout />
```
and

`PageLayout.tsx`

```
<Layout>
  <Page>
    {children}
  </Page>
</Layout>
```
The reason for this is partly because then DRB isn't dependent upon the ids `nypl-header` and `nypl-footer` which probably aren't likely to change, but could, and if they did, we would need to fix this issue again.

jackiequach:
@samanthaandrews The method you mentioned was used prior to the update to using the new header app. Due to how embeddable scripts work with Next.js, the header and footer need to be placed in the _document.tsx file. If not, they will disappear when navigating the app. I could not find a way to conditionally render/execute the header/footer components in the _document file.

samanthaandrews:
@jackiequach Thank you for the extra context. I understand now! I will approve this PR.

samanthaandrews (approved):

PR #417, opened 2023-06-23
SFR-1688: Next.js 13 upgrade

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1688)

- Upgrade to Next.js 13.4.7
- Fix linting errors from the addition of the new next eslint plugin
- Remove reference of @nypl/dgx-header-component

-


.eslintrc.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
   ],
   extends: [
     "eslint:recommended",
+    "next",
```
samanthaandrews, line 26: nice! would probably be good to make a separate ticket to clean up which of these eslint plugins and extensions are actually necessary.

samanthaandrews (approved):
Awesome work!

PR #437, opened 2023-09-06
SFR-1689: Upgrade testing library to v14

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1689)

- Upgrade testing-library packages to version 14 and jest to version 29.
- Update failing tests for async userEvent calls and update to default mockRouter path "/"

-


src/__tests__/componentHelpers/FilterYears.tsx
```diff
@@ -39,7 +40,7 @@ export const FilterYearsTests = (
       });
       fireEvent.change(fromInput, { target: { value: 1990 } });
       expect(fromInput).toHaveValue(1990);
-      userEvent.click(applyButton);
+      await user.click(applyButton);
```
samanthaandrews, line 43: It looks like there are two different user event actions we're using across all of these tests:

1. `fireEvent`
3. `userEvent`

It looks like rtl recommends using `userEvent`. Let's aim to replace `fireEvent` with `userEvent`, either in this PR or a follow-up one.

samanthaandrews (approved):
Since we have time, I think it would be worthwhile to create a follow-up PR to replace all `fireEvent` calls with `userEvent` since that's what the RTL docs recommend using. I'm ok with merging this PR as it is, but expect to see the follow-up in a few days. If you want to create a ticket to track that work you definitely can!

PR #442, opened 2023-09-08
Small adjustments to Playwright tests

samanthaandrews:
- Corrects catalog username and pin env var properties in `Playwright.yml`
- Careers page heading is now an `h2` 🤷 My recommendation is actually _not_ to test pages outside of DRB. Since we don't control the Careers page, we're setting ourselves up for flaky tests by making assertions on any page that we don't control. We can test that the link in our header exists, but actually clicking it doesn't necessarily need to happen. I don't think anything needs to be changed but just a note for the future.
- Ensures CI tests are being run against the QA URL, not Prod. We should never be running tests like this on Prod because it will skew our production Adobe Analytics and New Relic data. In the end this was probably why many of these tests were failing in CI but not locally. Locally, the QA URL is being set to `"https://drb-qa.nypl.org"` in the `cucumber.json` file. Which is why removing it from the `Playwright.yml` file fixes this problem. Now the tests are using the value from the `cucumber.json` file.
- Increased the global Cucumber timeout

My recommendation for updating Playwright tests is to _always_ run them in CI until they pass before making a PR ready for review. CI often behaves differently than locally, and the only way to know if we've fixed the issues is to actually run them in the CI env. To do this, I edit the `Playwright.yml` file with the following:

```
on:
  push:
    branches:
      - fix-flaky-playwright-tests
  # schedule:
  #   - cron: "00 14 * * 1-5"
  # pull_request:
  #   branches: [production]
  #   types: [opened, synchronize, reopened]
```

The branch name is the branch I'm currently using. I reverted this change before actually merging the PR, but if you check the previous commits you will see how I did this. You can see the successful run here https://github.com/NYPL/sfr-bookfinder-front-end/actions/runs/6123594835/workflow


samanthaandrews:
Let's wait to merge this until after the 0.17.4 prod release on Monday. I'll move it out of draft state and make it ready for review when that happens. But for now go ahead and read over the description and see the updates I made.

ayan1229 (approved):

clarissarichard (approved):

PR #443, opened 2023-09-08
SFR-1779: Fix incorrect prod args

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1779)

- Update the production yaml file to have a unique ECR image tag.

When our ECS cluster automatically restarts, the :latest docker build is pulled, which could be a QA build. This was never an issue since we had not used hard-coded build arguments prior to the New Relic instrumentation.


samanthaandrews (commented):
I am thinking we should make the same change to the QA build file. Also, we should aim to get this into the 0.17.4 release to ensure it's fixed early next week.

samanthaandrews (approved):
I knew _somehow_ our deployment was grabbing the wrong image. Good find!!

PR #444, opened 2023-09-14
SFR-1721: Remove TemplateFooter and upgrade DS

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1721)

- Remove TemplateFooter empty components
- Upgrade to DS 1.7.3, added workarounds for Hero `backgroundColor` and `foregroundColor` [issues](https://jira.nypl.org/browse/DSD-1281).  

-


samanthaandrews (changes requested):
This does what the ticket describes, but since this is a dev-led ticket, I'd really like to see some additional optimizations.

The ultimate goal for this issue is to reduce the number of times we're repeating components in DRB. I can see that we repeat components like `<Template />` `<TemplateContent />` `<TemplatePrimary />` and more on most pages throughout DRB. In this PR I'd like to see a solution that removes these repetitions, perhaps by creating a parent component that wraps these pages? Something like...

```
const DRBTemplate: FC = ({breadcrumbsProps}) => (
    <Template>
      <TemplateBreakout>
        <Breadcrumbs
          breadcrumbsType={breadcrumbsProps.type}
          breadcrumbsData={breadCrumbsProps.otherData}
        />
    // might need to pass this in as a prop since sometimes it could be the search component instead of the DrbHero
        <DrbHero />
      </TemplateBreakout>
      <TemplateContent>
        {children}
      <TemplateContent/>
    </Template>
)
```

samanthaandrews (changes requested):
This is great, a big improvement. I think we can take it one step further to make this even cleaner.

What if we put the `<DRBTemplate />` component into the `<Layout />` component? Or, put it in each of the pages where the <Layout /> component is rendered? I think the second options is a little cleaner because when you go to the AdvancedSearchPage, for example, you'd see...

```
<Layout>
  <DRBTemplate
    breadcrumbsData={breadcrumbsData}
    contentTop={contentTopElement}
  >
    <AdvancedSearch languages={languages} />
  </DRBTemplate>
</Layout>
```

This means that all of the complex AdvancedSearch logic is still inside that component, but all of the header and breadcrumbs stuff is pulled out into a wrapper which is a good separation of concerns.

I also think that instead of passing in `primaryContent` as a prop, that should be rendered as {children} inside the `<DRBTemplate />` component, because every component will render primary content.

Let me know what you think. Happy to hear additional improvements that you think could be made to the composition of these components.

src/components/About/About.tsx
```diff
@@ -1,128 +1,108 @@
...
-            missing, if you find an error please tell us in the feedback!
-          </p>
-        </TemplateContentPrimary>
-      </TemplateContent>
-      <TemplateFooter />
-    </Template>
+const About: React.FC = () => {
+  return (
+    <DrbTemplate
+      breadcrumbsDataProps={[{ url: "/about", text: "About" }]}
```
samanthaandrews, line 103: let's just call this `breadcrumbsData` if possible. the props is a little redundant i'm realizing (my bad on the naming here).

samanthaandrews:
Ok, I understand and agree that pulling out the state doesn't make this cleaner.

I do think there is a way to compose this a little better that I didn't think of before, but has become clearer to me this morning. I hesitate to use props to pass these components because to me, props are better used for passing data and other state values. For composing components, I'd rather just render the components directly.

What do you think of something like this instead?

```
const Collection = () => (
  <DRBTemplate>
  // could name this `<DRBBreadcrumbs /> or something else?? whichever names for these components make the most sense to you seems fine to me
  <Breadcrumbs breadcrumbsData={breadcrumbsData} />
  <ContentTop>
    {/* add the top content children here */}
  </ContentTop>
  <Header>
  {/* add the header content children here */}
  </Header>
  <Collection>
  {/* add the collection children here */}
  </Collection>
</DRBTemplate>
)
```

Breadcrumbs, ContentTop, and Header would all be new components that you would create. These components would wrap the shared Reservoir components and styles. Breadcrumbs, ContentTop, Header, and DRBTemplate would all render `{children}`.

This is only a slight change from what we have, but I do think it's an improvement since we don't have to check for whether or not components exist as props in the DRBTemplate. Let me know what you think and thanks for your patience as I work through this.

jackiequach:
I updated DrbTemplate to render {children} and added the appropriate supporting components. In order for the children of DrbTemplate to render under the correct DS components, DrbBreakout, DrbContent, DrbContentTop, and DrbContentPrimary were added. I could not figure out a way to get around this without passing elements through props. 

Alternatively, there is [TemplateAppContainer](https://nypl.github.io/nypl-design-system/reservoir/v1/iframe.html?viewMode=docs&id=components-page-layout-template--docs&args=#templateappcontainer-overview) from the DS which removes the need to import all the Template components individually.

samanthaandrews:
> Alternatively, there is [TemplateAppContainer](https://nypl.github.io/nypl-design-system/reservoir/v1/iframe.html?viewMode=docs&id=components-page-layout-template--docs&args=#templateappcontainer-overview) from the DS which removes the need to import all the Template components individually.

Wow I had no idea this existed! This is a great option. I'm ok with what you have now OR using this option, since it is recommended by the DS. Totally up to you.

samanthaandrews (approved):

PR #445, opened 2023-09-26
SFR-1780: Replace fireEvent with userEvent

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1780)

- Replace fireEvent with userEvent

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix: Small adjustments to Playwright tests
 - Upgrade testing-library packages to v14 and jest to v29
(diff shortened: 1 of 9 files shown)


samanthaandrews (approved):
This looks great!

Something I noticed is that there is a custom render function that we're using in `Search.test.tsx` and `Collection.test.tsx`, but not using in our other tests. I think we should always be using the custom render function, since this is what RTL recommends in their docs, https://testing-library.com/docs/react-testing-library/setup#custom-render. Can you create another ticket for this change?

PR #446, opened 2023-09-27
SFR-1721: Fix issues from QA

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1721)

- Fixes the issue with the Search Results page primary content not being positioned correctly with the sidebar. 
    - Due to how Chakra styles are inherited, the custom DrbTemplate component does not pass the variant to the primary content component.

-


samanthaandrews (approved):
I like the use of `<TemplateAppContainter />` 👍

PR #470, opened 2023-12-08
Update Node to v18.18.2

ayan1229:

- updates/changes the node version to 18.18.2 to address playwright test failure issues


package.json
```diff
@@ -15,7 +15,7 @@
   },
   "dependencies": {
     "@chakra-ui/react": "2.5.4",
-    "@newrelic/next": "^0.7.0",
+    "@newrelic/next": "^0.6.0",
```
jackiequach, line 18: The version for these packages shouldn't need to be changed for this PR.

jackiequach (changes requested):

package.json
ayan1229: @jackiequach undid the changes. 

Thanks

samanthaandrews (approved):

PR #471, opened 2023-12-11
SFR-1854: Add physical edition badge

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1854)

- Add physical edition badge and Scan and Deliver blurb to EDD editions
- Update styles to match designs
    - Card actions width to 200px
    - Login button is secondary

-


src/components/InstanceCard/InstanceCard.tsx
```diff
@@ -56,6 +67,7 @@ export const InstanceCard: React.FC<{
         </div>
         <div>{EditionCardUtils.getWorldCatElem(instance)}</div>
         <Link to="/license">{EditionCardUtils.getLicense(previewItem)}</Link>
+        {isPhysicalEdition && EditionCardUtils.getScanAndDeliverBlurb()}
```
samanthaandrews, line 70: I know this is a pattern we've been using throughout DRB, but I recommend refactoring this to explicitly render a component instead of a function that returns JSX. This makes our components more scalable for the future, and it is also a slight performance improvement.

`{isPhysicalEdition && <ScanAndDeliverBlurb />}`

For this PR, let's ensure all new components are using this method. If there are other instances throughout the code where we are returning JSX from a function instead of rendering the component directly, let's make a ticket to update that so that it matches the recommended pattern.

samanthaandrews (changes requested):

jackiequach:
Created a ticket for updating the rest of edition utils to use the mentioned pattern: https://jira.nypl.org/browse/SFR-1860

Switched the base branch to limited-access feature branch since there will be multiple PRs related to the work.

samanthaandrews (approved):

PR #479, opened 2024-02-09
SFR-1871: Implement authorization for UP download requests

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1871)

- Move login prop to CTA component to reduce prop inheritance 
- Add API implementation for UP download fulfillment endpoint

-

### DIFF
--- a/src/components/EditionCard/DownloadLink.tsx
+++ b/src/components/EditionCard/DownloadLink.tsx
@@ -2,6 +2,7 @@ import { Box, Icon } from "@nypl/design-system-react-components";
 import React from "react";
 import Link from "~/src/components/Link/Link";
 import { trackCtaClick } from "~/src/lib/adobe/Analytics";
+import { fulfillFetcher } from "~/src/lib/api/SearchApi";
 import { ItemLink } from "~/src/types/DataModel";
 import { formatUrl } from "~/src/util/Util";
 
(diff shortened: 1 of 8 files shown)


src/lib/api/SearchApi.ts
```diff
@@ -120,3 +120,18 @@ export const readFetcher = async (linkId: number) => {
...
+    method: "GET",
+    headers: {
+      Authorization: `Bearer ${nyplIdentityCookie.access_token}`,
+    },
+  });
+  if (res.ok) {
+    return res.url;
+  } else {
+    throw new Error(`Unable to download UP PDF`);
+  }
```
samanthaandrews, line 136: I'm thinking this might be a good opportunity to sync with Apoorva on the potential errors that could come up here, based on what the server responds with, and checking to see if she has input on if we want to show anything to the user.

Based on her input, we should write a test to make sure the error gets shown in the way we expect it to.

src/components/EditionCard/DownloadLink.tsx
```diff
@@ -1,20 +1,29 @@
...
-}> = ({ downloadLink, title, isLoggedIn }) => {
+  loginCookie?: string;
+}> = ({ downloadLink, title, isLoggedIn, loginCookie }) => {
+  const router = useRouter();
   if (downloadLink && downloadLink.url) {
     let linkText = "Download PDF";
     let linkUrl = formatUrl(downloadLink.url);
 
-    const trackDownloadCta = () => {
+    const handleOnClick = async (e) => {
```
samanthaandrews, line 21: Let's rename this to something more descriptive like `handleDownload` or `handleFulfillmentDownload` -- not sure what makes the most sense here.

samanthaandrews: ohh i see this is used for multiple different types of downloading/viewing content. so maybe `handleDownload` is best? or maybe `handleOnClick` is ok since it is generic.

samanthaandrews (changes requested):
I made a couple of comments about checking with Apoorva on showing errors to users.

A few other comments:

- Do we have a test for making sure the cookie gets added as expected after logging into nypl.org?
- Do we have a test for making sure the link with the `/fulfill/` path fetches the resource as expected and redirects a user where they are supposed to go?

I realize these tests may be difficult to implement with Jest alone. An option would be to implement Playwright for these tests, which we would need a follow up ticket to do.

src/lib/api/SearchApi.ts
jackiequach: Working with Apoorva to get designs approved for a modal which would appear for servers errors unrelated to authenication. For authentication errors, it will redirect to the login page.

samanthaandrews (approved):

PR #486, opened 2024-03-15
SFR-1898: Combine publisher lists on edition page

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1898)

- Combine publisher lists to be a single list for better accessibility

-


oliviawongnyc (approved):
Makes sense to me!

PR #487, opened 2024-03-19
SFR-1897: Add Playwright integration tests

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1897)

- Setup Playwright configuration for integration tests
- Add integration tests for downloading limited access books
- Update CI.yml to include Playwright integration tests and update action versions
- Update Dockerfile to follow Next.js example and copy next.config.js to ignore msw during build

-


.gitignore
```diff
@@ -23,4 +23,8 @@ vega.spec.ts
 newrelic_agent.log
 
 .vscode
-.vscode/launch.json
\ No newline at end of file
+.vscode/launch.json
+
+# playwright
+playwright-report
```
kylevillegas93, line 29: let's add a new line!

Dockerfile
```diff
@@ -1,42 +1,48 @@
...
-EXPOSE $PORT
+COPY --from=builder /app/next.config.js ./
+COPY --from=builder /app/public ./public
 
-# Build the app!
-RUN npm run build
+EXPOSE $PORT
 
 # CMD is the default command when running the docker container.
 CMD npm run start:newrelic
```
kylevillegas93, line 48: Let's add a new line here at the bottom !

next.config.js
```diff
@@ -1,9 +1,30 @@
...
-module.exports = {
-  webpack: (config) => {
+/**
+ * @type {import('next').NextConfig}
+ */
+const nextConfig = {
+  webpack: (config, { isServer }) => {
     config.resolve.alias["~"] = path.resolve(__dirname);
 
+    // should only include these for testing env
```
kylevillegas93, line 10: I think the if condition is clear that these config values are only being included for testing. Feel free to remove the comment.

playwright/integration/auth.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  test("redirects to NYPL log in page with no cookie", async ({
+    page,
+    port,
+  }) => {
+    await page.goto(`http://localhost:${port}${LIMITED_ACCESS_WORK_PATH}`);
+    await page.getByRole("link", { name: "title Download PDF" }).click();
+    await page.waitForURL(`**${NYPL_LOGIN_URL}**`);
+    const url = new URL(page.url());
+    const redirectUri = url.searchParams.get("redirect_uri");
+    expect(redirectUri).toContain(LIMITED_ACCESS_WORK_PATH);
```
kylevillegas93, line 29: A suggestion here to make this super clear is to add a new line between setup and the expectation/assertion

kylevillegas93 (approved):
Overall looks good! Couple minor non-blocking comments!

mocks/index.ts
```diff
@@ -0,0 +1,15 @@
+async function initMocks() {
+  if (typeof window === "undefined") {
+    const { server } = await import("./server");
+    server.listen({
+      onUnhandledRequest: "bypass",
```
Toxiapo, line 5: Should we add this line to the browser worker too?

playwright/integration/auth.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  LIMITED_ACCESS_WORK_PATH,
+  NYPL_LOGIN_URL,
+} from "~/mocks/mockEnv";
+import { server } from "~/mocks/server";
+
+test.beforeEach(async ({ context }) => {
+  await context.clearCookies();
+  server.listen({
+    onUnhandledRequest: "bypass",
+  });
```
Toxiapo, line 14: I am curious why we need to initialize this when we already do that in the `_app.tsx` file.

jackiequach: This is not needed here, it was added when I was initially implementing this. Nice catch

Toxiapo: Gotcha! I think if we remove this, then we might need to also remove the `server.close()` and `server.resetHandlers` lines below, right? otherwise new tests will start without msw.

I think you wanted to entirely rely on the MSW server when running the app for the tests so it should always stay on. Unless I am not understanding this correctly.

jackiequach: A new server is started for each test using the `port` fixture, so resetting the handlers ensures the data is clean.  Closing the server after all tests are run is to clean the test env.

Toxiapo (approved):

PR #491, opened 2024-03-26
SFR-1905: Add build arg to yml files

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1905)

- Previously had added the env var for APP_ENV to the yml build files, but did not include it as a build-arg for the docker build command.

-


samanthaandrews (approved):
Looks good to me. Wondering if you were able to test this locally to make sure data is going through to NR with this dockerfile update?

jackiequach:
Updated dockerfile with env var after testing locally. It works correctly now.

PR #498, opened 2024-05-03
SFR-1929: Fix 0 character issue and misc performance improvements

jackiequach:
[Jira Ticket](http://jira.nypl.org/browse/SFR-1929)

- Fixes issue where 0 character is shown when there are no authors
- Replace HTML scripts with Next.js Script and add styles to header placeholder to prevent content shift on page load

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Update Github Action for release from create-release to action-gh-release
 - SFR-1945: Fixed more failing Playwright tests
 - Update README to include production release information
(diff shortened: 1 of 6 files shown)


Toxiapo (approved):
The update makes sense to me! However, I am unable to open the preview link. Does it work for you?

jackiequach:
@Toxiapo The Vercel preview does not work since DRB is behind VPN and to get it working we would need to upgrade to Vercel Enterprise.

kristojorg (approved):

PR #508, opened 2024-07-12
SFR-2033/verify-nypl-header-links

ayan1229:
[Jira Ticket](https://newyorkpubliclibrary.atlassian.net/browse/SFR-2033)

- verifies the external NYPL app links nested inside DRB app
- verifies when clicked on the links user is directed to the correct landing pages.

### DIFF
--- a/playwright/features/headerLinks.feature
+++ b/playwright/features/headerLinks.feature
@@ -18,17 +18,17 @@ Feature: Header Links
         And the "get help header link" should be displayed
         And the "search header link" should be displayed
 
-           Examples:
-            | DRB                 |
-            | "home"              |
-            | "advanced search"   |
-            | "search results"    |
-            | "item details"      |
-            | "edition details"   |
-            | "collection"        |
-            | "read online"       |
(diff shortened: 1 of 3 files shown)


playwright/features/headerLinks.feature
```diff
@@ -18,32 +18,45 @@ Feature: Header Links
...
+            | "search header link"     | "search header button"                   |
+            | "search header link"     | "close search header link"               |
+
+    Scenario: As a user when I click on Locations I should be directed to NYPL location page
+        Given I go to the "home" page
+        Then I click the <first header link>
+        And the <landing page header> should be displayed
+
+        Examples:
+            | first header link           | landing page header              |
```
Toxiapo, line 57: I am not familiar with cucumber syntax. Does it always start with a row without double quotes?

ayan1229: @Toxiapo yes. First row will always be without quotes to map the next rows with the main steps

Toxiapo (approved):

PR #509, opened 2024-07-25
SFR-2056: Update proxyUrl for limited access books

jackiequach:
[Jira Ticket](https://newyorkpubliclibrary.atlassian.net/browse/SFR-2056)

- Only pass the proxyUrl to the web reader when the resource is not internally hosted (limited access books)

-


kylevillegas93 (approved):
Looks good! Are there any other cases where we don't necessarily need to use the proxy?

jackiequach:
@kylevillegas93 The proxy isn't used when ePubs/html resources are passed, but that is already handled on the web reader side. I don't believe we host any resources other than limited access.

PR #522, opened 2024-08-23
SFR-2136: Implement error boundary

jackiequach:
[Jira Ticket](https://newyorkpubliclibrary.atlassian.net/browse/SFR-2136)

- Implements error boundary for uncaught exceptions

-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix error when a collection is empty
 - SFR-2077: Validate top 5 language filters for Frontend DRB search results
 - Add Playwright integration test for cookie validation
+- Add Error Boundary for uncaught exceptions
 
 ## [0.18.2]
 
--- a/mocks/handlers.ts
+++ b/mocks/handlers.ts
@@ -1,7 +1,13 @@
 import { http, HttpResponse, passthrough } from "msw";
+import {
+  collectionItem,
+  oneCollectionListData,
+} from "~/src/__tests__/fixtures/CollectionFixture";
 import { workDetailWithUp } from "~/src/__tests__/fixtures/WorkDetailFixture";
 import {
   API_URL,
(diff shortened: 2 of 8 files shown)


kylevillegas93:
The change looks good. Should the error boundary work for all uncaught exceptions? When running locally and navigating to http://localhost:3000/edition/1 I see the following:

```
Server Error
Error: cannot find work with identifier 1

This error happened while generating the page. Any console logs will be displayed in the terminal window.
```

kylevillegas93:
After running npm i; npm run build; and npm run start everything looks good!

kylevillegas93 (approved):

Toxiapo:
> The change looks good. Should the error boundary work for all uncaught exceptions? When running locally and navigating to http://localhost:3000/edition/1 I see the following:
> 
> ```
> Server Error
> Error: cannot find work with identifier 1
> 
> This error happened while generating the page. Any console logs will be displayed in the terminal window.
> ```

@kylevillegas93 React's error boundary only catches errors during component renders, things like a failed fetch request or syntax errors within the component itself, and doesn't catch routing level errors so this should be fine.

However, I believe this could still be a bug where an invalid URL should show the 404 page in the pages/404.tsx instead of a generic 404 page from the browser, what do you think? @jackiequach

jackiequach:
@Toxiapo Kyle and I spoke offline about this. The API files need to be updated to properly throw the correct status code and will be addressed in the future.

Toxiapo (approved):
LGTM!

Toxiapo:
@jackiequach, this isn't directly related to this PR, but I noticed a small UI bug. When searching for a book, the skeleton loader seems to cover the whole screen, which makes it look like the page is broken. Thought I'd mention it!

Screenshot 2024-08-29 at 10 44 44 AM

PR #551, opened 2024-11-04
NO-REF: Fix header when banner is up

kylevillegas93:
[Jira Ticket](https://newyorkpubliclibrary.atlassian.net/browse/SFR-XXX)

- Fixes header when banner is up

-  npm run dev

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGE LOG
 
 ## [Prerelease]
+- Fix header when banner is up
 
 ## [0.18.6]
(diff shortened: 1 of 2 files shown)


ayan1229 (approved):

samanthaandrews (approved):
I think this is ok as a quick fix for now but I wonder if we need to add this to the header repo so that everyone using the header would benefit from this fix. Adding style straight to the div gives me pause. I'll discuss with @jackiequach when she's back.

kylevillegas93:
@samanthaandrews ya not super ideal :|. these script based components are a bit tricky to work with

PR #554, opened 2024-11-05
SFR-2078: Automate Different Publishing Sources

ayan1229:
[Jira Ticket](https://newyorkpubliclibrary.atlassian.net/browse/SFR-2078)

- validates following collections from different sources 

OAPEN PDF: How safe is eating chicken? 

Gutenberg ePub: Chicken Farm 

Project Muse PDF: Reading fiction in antebellum America 

IntechOpen PDF: Climate Change and Agriculture

### DIFF
--- a/cucumber.json
+++ b/cucumber.json
@@ -2,7 +2,7 @@
   "default": {
     "import": ["playwright/tests/**/*.ts", "playwright/support/**/*.ts"],
     "worldParameters": {
-      "appUrl": "https://drb-qa.nypl.org/",
+      "appUrl": "http://localhost:3000/",
       "headless": false
     }
   }
--- a/cucumber.json
+++ b/cucumber.json
@@ -2,7 +2,7 @@
   "default": {
     "import": ["playwright/tests/**/*.ts", "playwright/support/**/*.ts"],
     "worldParameters": {
-      "appUrl": "http://localhost:3000/",
+      "appUrl": "https://drb-qa.nypl.org/",
       "headless": false
     }
   }
--- a/playwright/features/pubSource.feature
+++ b/playwright/features/pubSource.feature
@@ -1,0 +1,32 @@
+Feature: Specific Pub Sources
+    Scenario: As a user I search for title "How safe is eating chicken" and open the title after search result is loaded
+        Given I go to the "home" page
+        When I select "title" from the category dropdown
+        And I fill in the "homepage search box" with "How safe is eating chicken?"
+        And I click the "search button"
+        And I click the "first read online button"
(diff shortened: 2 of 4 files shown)


cucumber.json
```diff
@@ -2,7 +2,7 @@
   "default": {
     "import": ["playwright/tests/**/*.ts", "playwright/support/**/*.ts"],
     "worldParameters": {
-      "appUrl": "http://localhost:3000/",
+      "appUrl": "https://drb-qa.nypl.org/",
```
samanthaandrews, line 5: Should this be reverted to localhost or are we going to hit the live QA site?

ayan1229: We will do it on localhost once the VPN is removed for API. 
Kathleen approved it. We are just waiting for it to be removed. Then we will run it on the localhost

jackiequach: Not sure if I am missing some context from when I was out, but wouldn't this be the opposite since QA is behind the VPN and localhost would allow you to run the tests outside of the time window QA is available? Once the VPN is removed we should run these tests on the QA env.

kylevillegas93: Yes I think that the tests should run against localhost

playwright/features/pubSource.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Specific Pub Sources
```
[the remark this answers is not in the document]
ayan1229: This is cucumber test scenarios. Basically all these steps are being tested and test steps are dynamically written in our test folder. Cucumber is used for easier documentation and quick test writing.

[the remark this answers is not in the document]
kylevillegas93: Ok thanks! I understand. 

These test scenarios are brittle and verbose.

I recommend considering breaking out the specific functionality to test: 
- Search functionality - ensure search results show up as expected given a very generic query
- Reader functionality - ensure a specific file/type or source is readable.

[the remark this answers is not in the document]
samanthaandrews: Since we've added Playwright & Cucumber to the app, I've learned a lot more about how to structure Playwright tests, and the general best practice is to use a [page object model](https://playwright.dev/docs/pom). As an example, you can see how we are doing this in [pjr](https://github.com/NYPL/ereading-clients/tree/main/apps/pjr/playwright/discover): each page has its own folder that contains a spec file and a corresponding page file.

Also, I'd like to re-evaluate if we need to use Cucumber at all. Imo the layer of abstraction it provides doesn't provide any value and adds unnecessary complexity.

My recommendation would be to remove Cucumber and begin refactoring our tests to use the page object model.

PR #569, opened 2024-12-18
SFR-2416: Run e2e tests on local domain

jackiequach:
[Jira Ticket](https://newyorkpubliclibrary.atlassian.net/browse/SFR-2416)

- Updates GH action to run e2e tests on local.nypl.org for PRs not merging production. 
- e2e tests will run on drb-qa.nypl.org for production PRs.
- Re-enables the e2e test for login 

-


playwright/integration/auth.spec.ts
```diff
@@ -14,7 +14,7 @@ test.beforeEach(async ({ context }) => {
 test.afterEach(() => server.resetHandlers());
 test.afterAll(() => server.close());
 
-test.describe("Cookie authentication", () => {
+test.skip("Cookie authentication", () => {
```
kylevillegas93, line 17: does this test still not pass? should we skip?

.github/workflows/Playwright.yml
```diff
@@ -15,6 +15,19 @@ jobs:
         with:
           node-version-file: ".nvmrc"
 
+      - name: Set environment for branch
+        run: |
+          if [[ $GITHUB_REF_NAME == 'production' ]]; then
```
kylevillegas93, line 20: should this be `[[ $GITHUB_REF_NAME == 'development' ]]`?

samanthaandrews: since we talked about getting rid of the produciton/development branches and just using main for full CI/CD i'm wondering if this makes sense to use as a check. is there another check we could use?

playwright/integration/auth.spec.ts
jackiequach: The tests didn't pass due to a double redirect for the login.nypl.org page. I updated the tests to check for the final destination.

.github/workflows/Playwright.yml
kylevillegas93: definitely agree we should change the branch structure eventually. 

@jackiequach - just double checking on the logic here. Should we use drb-qa.nypl.org for production github ref?

jackiequach: These should run when there's a PR merging into the production branch, so before release.

kylevillegas93 (approved):

PR #585, opened 2025-01-13
NO-REF: Fix Playwright env setting step

jackiequach:
- Updates the Playwright.yml step for setting environment and calling the workflow from ci.yml 

-


kylevillegas93 (approved):
Looks good!

PR #586, opened 2025-03-24
NO-REF: Update to Next.js 13.5.9

jackiequach:
- Updates to Next.js 13.5.9
- Resolves some dependency conflicts
- Updates the EDD test to only check if the Request Scan button goes to the research catalog and not the text on the RC page

-


samanthaandrews (approved):
Code looks good just need to fix the failing CI checks. Thanks for doing this!

